PR #35, opened 2015-01-14
Dynamic accessing of children in-line with how Form does it.

Weeblewonder:
Form has a magic getter to get the fields it contains, but a child form does not even though its technically meant to be the same kind of entity.
Added this to my local install as it makes things more consistent, pushing up here to see if you agree.


kristijanhusak:
It's good idea, but i would rather put it into the `ParentType`, so other classes that have `children` can also use it. Just move the code there and it will be great.

Weeblewonder:
No worries, will make another PR.
Considered putting it in parent type but dont know enough about potential children of ParentType to know whether it would be equally applicable as I was going for consistency across form types.

PR #39, opened 2015-01-28
Readme suggestion causes exception

stevebauman:
Using a string inside the wrapper option causes an exception on line [171](https://github.com/kristijanhusak/laravel-form-builder/blob/laravel-4/src/Kris/LaravelFormBuilder/FormHelper.php#L171) on FormHelper due to it using an array and not a string.


kristijanhusak:
Thanks :)

PR #50, opened 2015-02-27
Don't override existing aliases for Form and Html

squigg:
Some other packages register the same `Form` or `Html` aliases, but use
different identifiers for the class within the IOC. This commit adds a
check for whether there is already an existing facade aliased so it does
not overwrite them

### DIFF
--- a/src/Kris/LaravelFormBuilder/FormBuilderServiceProvider.php
+++ b/src/Kris/LaravelFormBuilder/FormBuilderServiceProvider.php
@@ -77,10 +77,13 @@ class FormBuilderServiceProvider extends ServiceProvider
                 return $form->setSessionStore($app['session.store']);
             });
 
-            AliasLoader::getInstance()->alias(
-                'Form',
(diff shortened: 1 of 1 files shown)


kristijanhusak:
Can you give me the link to package where form and html are registered with different identifier from IOC?

I'm using Form facade in views, and if that facade is not pointing to the right class, it can cause some problems here. If the situation is like that i'll have to skip using facade and pass class directly from IOC.

squigg:
https://github.com/patricktalmadge/bootstrapper

It's registered as bootstrapper::form and has it's own Facade that points to that class instead.It just extends the default Laravel Form with some extra methods, so all your views still work fine.

I don't know of one that does it for Html, but thought better be safe.

There's an assumption that different implementations of these classes wouldn't reduce the default Laravel functionality. But suppose developers should know what they're doing combining multiple Form related packages.

kristijanhusak:
Ok great! Thanks for PR :)

PR #55, opened 2015-03-19
Set data for child forms and create named forms.

tormit:
Not sure if this was bug or missing feature, but I was not able to pass data for child forms. Now its possible.
Also added new method in FormBuilder to create named forms like Symfony does. Form field names become like youname[fieldname].

### DIFF
--- a/src/Kris/LaravelFormBuilder/FormBuilder.php
+++ b/src/Kris/LaravelFormBuilder/FormBuilder.php
@@ -54,6 +54,37 @@ class FormBuilder
     }
 
     /**
+     * Create named form to group fields in this form by a name.
+     * Fields are named like name[field].
+     *
+     * @param $name
+     * @param $formClass
+     * @param array $options
+     * @param array $data
+     * @return Form
+     */
+    public function createNamedForm($name, $formClass, array $options = [], array $data = [])
+    {
+        $dataReal = $data;
+        if (empty($dataReal)) {
+            if (isset($options['model']) && is_object($options['model']) && method_exists($options['model'], 'toArray')) {
+                $dataReal = $options['model']->toArray();
+            }
+        }
+
+        return $this->plain(
+            $options
+        )->add(
+            $name,
+            'form',
+            [
+                'class' => $this->create($formClass, $options, $dataReal),
+                'label' => isset($options['label']) ? $options['label'] : false
+            ]
+        );
+    }
+
+    /**
      * Get the namespace from the config
      *
      * @return string
--- a/src/Kris/LaravelFormBuilder/FormBuilder.php
+++ b/src/Kris/LaravelFormBuilder/FormBuilder.php
@@ -1,6 +1,7 @@
 <?php  namespace Kris\LaravelFormBuilder;
 
 use Illuminate\Contracts\Container\Container;
+use Illuminate\Database\Eloquent\Model;
 
 class FormBuilder
 {
@@ -41,6 +42,7 @@ class FormBuilder
             );
         }
 
+        /** @var Form $form */
         $form = $this->container
             ->make($class)
             ->setFormHelper($this->formHelper)
@@ -48,40 +50,18 @@ class FormBuilder
             ->setFormOptions($options)
             ->addData($data);
 
-        $form->buildForm();
-
-        return $form;
-    }
-
-    /**
-     * Create named form to group fields in this form by a name.
-     * Fields are named like name[field].
-     *
-     * @param $name
-     * @param $formClass
-     * @param array $options
-     * @param array $data
-     * @return Form
-     */
-    public function createNamedForm($name, $formClass, array $options = [], array $data = [])
-    {
-        $dataReal = $data;
-        if (empty($dataReal)) {
-            if (isset($options['model']) && is_object($options['model']) && method_exists($options['model'], 'toArray')) {
-                $dataReal = $options['model']->toArray();
+        // reset model to match named form logic
+        if ($form->getName()) {
+            if ($form->getModel() instanceof Model && method_exists($form->getModel(), 'toArray')) {
+                $form->setModel([$form->getName() => $form->getModel()->toArray()]);
+            } elseif (is_array($form->getModel())) {
+                $form->setModel([$form->getName() => $form->getModel()]);
             }
         }
 
-        return $this->plain(
-            $options
-        )->add(
-            $name,
-            'form',
-            [
-                'class' => $this->create($formClass, $options, $dataReal),
-                'label' => isset($options['label']) ? $options['label'] : false
-            ]
-        );
+        $form->buildForm();
+
+        return $form;
     }
 
     /**
--- a/src/Kris/LaravelFormBuilder/FormBuilder.php
+++ b/src/Kris/LaravelFormBuilder/FormBuilder.php
@@ -52,10 +52,15 @@ class FormBuilder
 
         // reset model to match named form logic
         if ($form->getName()) {
+            $modelData = [];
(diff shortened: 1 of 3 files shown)


kristijanhusak:
How were you trying to pass data to child forms? Can you give me some code examples?

Creating named forms is great idea, but it can be done in another way, much simpler.
If you check [here](https://github.com/kristijanhusak/laravel-form-builder/blob/master/src/Kris/LaravelFormBuilder/Fields/ChildFormType.php#L72), you will see that you just need to pass `is_child` flag, and name that form needs to carry. So basically this is enough:
```php
    $form = FormBuilder::plain([
        'is_child' => true,
        'name' => 'yourname',
        'model' => $model
    ])->add('name')->add('email', 'email');
```

and you get `yourname[name]` and `yourname[email]`

The only problem here is that if you pass in model, it's data needs to be structured in that way:
```php
$model = [
    'somefield' => 'sometext',
    'yourname' => [ 'name' => 'example name', 'email' => 'test@example.com']
];
```

In the background model binding is done by Laravel's form builder class, so this is something that i'm not sure can be fixed, maybe just to check if not passed like that, and then to wrap it.

tormit:
The problem with 
```php
   $form = FormBuilder::plain()
```
is that it creates only plain form, but I need to create named form from pre-defined form.
And the issue about passing to child form was related to this named forms case.

The idea of createNamedForm() is to be like it is in Symfony. 
```php
   // form builder in Symfony
   $form = $this->createForm(
                $this->get('my.form.service'),
                $entity,
                [
                    'attr' => []
                ]
            )
```
Symfony FormBuilder creates named forms out of box.

Another possibility to make your FormBuilder similar to Symfony is to enable user to implement _getName()_ method in form definition class. If user has implemented _getName()_, then _FormBuilder::create()_ automatically created named form.

kristijanhusak:
You can do same thing with existing form:
```php
namespace App\Forms;

class MyForm extends Form {
    public function buildForm() 
    {
        $this->add('name')->add('email', 'email');
    }
}

// Controller
 $form = FormBuilder::create('App\Forms\MyForm', [
        'is_child' => true,
        'name' => 'yourname',
        'model' => $model
    ]);
```

Wrapper for this definitely needs to be created.

About `getName`, i would not implement that because there is already a property that can be set (`$childFormName`), but it will not work at this moment, so I need to fix that, and i'll probably skip `is_child` flag also. That way we will be able just to put property for example `$name` on a form, and any time it is created, it will add that.

kristijanhusak:
I just updated it to use only the `name` key. More info can be found [here](https://github.com/kristijanhusak/laravel-form-builder#named-forms). Form builder method for this is still needed, so fix your PR and i'll merge it.

tormit:
Cool. I'll do it tomorrow.

tormit:
Using you update, I'm able to create named forms, but passing existing data still won't work. My patch does not work anymore either. I'll try to solve it.

kristijanhusak:
Yeah, as i said, it expects data to be in that namespace, so if you pass `['name' => 'john', 'email' => 'johndoe@example.com']`, it needs to get it in this form:
```php
$data = [
    'form_name' => [
        'name' => 'johndoe',
        'email' => 'johndoe@example.com'
    ]
];
```

tormit:
I updated my comment with my code.

kristijanhusak:
3rd parameter for `create` method is used for passing data to the form class, but that data will not be used in the form itself. For that model is needed. Here's short explanation of using data:

```php
$data = ['some' => 'data passed for usage'];

$model = ['name' => 'johndoe', 'email' => 'johndoe@example.com'];

   $form = $formBuilder->create('\App\Forms\Customer\CompanyProfileForm', [
            'name' => 'profile',
            'method' => 'POST',
            'url' => route('customer.company.profile'),
            'model' => $model
        ], $data);


class CompanyProfileForm extends Form {
    public function buildForm()
    {
        $data = $this->getData('some'); // 'data passed for usage'
        $model = $this->getModel(); // Passed model, will be bound to form with laravels Form::model()
        $this->add('name')->add('email');
    }
}
```

This way it should work:
```php
$model = ['name' => 'johndoe', 'email' => 'johndoe@example.com'];

   $form = $formBuilder->create('\App\Forms\Customer\CompanyProfileForm', [
            'name' => 'profile',
            'method' => 'POST',
            'url' => route('customer.company.profile'),
            'model' => ['profile' => $model]
        ], $data);
```
Only problem is if the data is already namespaced in that way, there will be a problem. So some kind of check needs to be done.

tormit:
OK. Now I understand difference between data and model. 
The model object is present in CompanyProfileForm::buildForm(), but still no data is bound to form.

kristijanhusak:
Can you please give me the dump of data from the model that is passed?

tormit:
http://postimg.org/image/nt6x38n5v/e8ac4762/

kristijanhusak:
Thanks. As i see, you do not have any attributes with 'profile' key. So try doing this:
```php
$form = $formBuilder->create('\App\Forms\Customer\CompanyProfileForm', [
            'name' => 'profile',
            'method' => 'POST',
            'url' => route('customer.company.profile'),
            'model' => ['profile' => $model->toArray()]
        ], $data);
```

tormit:
Now it works. But I think it should be simpler. A developer would assume that given model automatically applies to the named form.

kristijanhusak:
Yeah, i know. That needs to be handled through the FormBuilder named form method that you want to add. Basically it should be enough to check if the model is passed, and if it's eloquent, to convert it to array, and then namespace it by the provided name, like you done it when passing the model.

tormit:
Check my update. I removed createNamedForm() as it is irrelevant now. Also added way to "fix" model for named forms.

src/Kris/LaravelFormBuilder/FormBuilder.php
```diff
@@ -48,6 +50,15 @@ public function create($formClass, array $options = [], array $data = [])
             ->setFormOptions($options)
             ->addData($data);
 
+        // reset model to match named form logic
+        if ($form->getName()) {
```
kristijanhusak: This logic can be moved maybe [here](https://github.com/kristijanhusak/laravel-form-builder/blob/master/src/Kris/LaravelFormBuilder/Form.php#L543).

src/Kris/LaravelFormBuilder/FormBuilder.php
```diff
@@ -48,6 +50,15 @@ public function create($formClass, array $options = [], array $data = [])
...
+            if ($form->getModel() instanceof Model && method_exists($form->getModel(), 'toArray')) {
+                $form->setModel([$form->getName() => $form->getModel()->toArray()]);
+            } elseif (is_array($form->getModel())) {
```
kristijanhusak: We also need to check if model maybe already have that namespace, so we do not double it.

tormit: Ok. I'll add fix soon.

tormit:
I made the fix. But I didn't understand how you want this logi to be moved.
You may take my changes and place this logic somewhere you consider more suitable.

kristijanhusak:
Ok, I will. Thanks for the help and idea for this!

tormit:
I see you have implemented model binding in your code. I assume that I may switch my project back to your version?
I'm closing this pull request.

kristijanhusak:
@tormit 
Yes, i did, but still didn't tag a release, want to add few more small things. Will tag it today probably.

tormit:
Ok. Thanks. This is great plugin. This is a missing feature of Laravel for those, like me, coming from Symfony.
Keep up good work!

kristijanhusak:
It's done, do a composer update :)

PR #76, opened 2015-04-17
Add ability to get all of $data from the form.

theshaunwalker:
I saw #55 after searching for issues and PR's to do with `getData()` but didnt seem to fit my issue as I dont use models with most of my forms.

What I'm looking for is flexibility to be able to get the the whole `$data` property of a form, rather than just 1 key which is all that is supported with the current `getData()` function.

### DIFF
--- a/src/Kris/LaravelFormBuilder/Form.php
+++ b/src/Kris/LaravelFormBuilder/Form.php
@@ -152,7 +152,7 @@ class Form
      * @return Form
      */
     public function modify($name, $type = 'text', array $options = [], $overwriteOptions = false)
-    {
+    {g
         // If we don't want to overwrite options, we merge them with old options
         if ($overwriteOptions === false && $this->has($name)) {
             $options = $this->formHelper->mergeOptions(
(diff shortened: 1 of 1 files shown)


kristijanhusak:
Yeah this is good idea, haven't thought about at all. Just fix that syntax error and i'll merge it :-) Thanks!

theshaunwalker:
Yeah, thanks Github's online file editor ¬_¬ lol

theshaunwalker:
Curious on your thoughts for something along the same problem I wanted this flexibility for.

I wanted it as a simpler way to pass data down to child forms, this way I just pass 
```
[
  'data' => $this->getData()
]
```
to child forms and I just manage the data in the root parent form.

To extend this it would be great if data was automatically passed to child forms when they're added, so in the `$form->add('foo', 'form)` function, when it's a `form` type it will automatically pass `$this->data` into it. This could be overriden if any data is passed in the actual function call so it doesnt remove how it currently works.

What do you think? I'll send another PR in if this is something you agree with.

kristijanhusak:
It's good idea, feel free to make PR :) Thanks for this one.

PR #77, opened 2015-04-18
Add ability to exclude fields from rendering.

theshaunwalker:
Adds logic to `Form`, `ChildFormType` and respective views, to add the ability to mark fields to not render when calling `->renderForm()` or `->render()`.

When using with `Form`:
```
$form->exclude(['unwanted_field', 'another_field'])->renderForm();
```

When using with `ChildFormType`:
```
$form->child->render(
  ['exclude' => 
    ['unwanted_field', 'another_field']
  ]);
```

Not experienced with phpunit for testing so havnt included tests as when I tried to copy another test and tweak to fit this use-case, Mockery was throwing stuff at me that I dont know how to solve :P

### DIFF
--- a/src/Kris/LaravelFormBuilder/Fields/ChildFormType.php
+++ b/src/Kris/LaravelFormBuilder/Fields/ChildFormType.php
@@ -47,7 +47,8 @@ class ChildFormType extends ParentType
             'class' => null,
             'default_value' => null,
             'formOptions' => [],
-            'data' => []
+            'data' => [],
+            'exclude' => []
         ];
     }
 
--- a/src/Kris/LaravelFormBuilder/Form.php
+++ b/src/Kris/LaravelFormBuilder/Form.php
@@ -61,6 +61,13 @@ class Form
     protected $formBuilder;
 
     /**
+     * List of fields to not render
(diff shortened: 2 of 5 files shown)


theshaunwalker:
Apologies for the multiple commits, forgot I should probably run phpunit locally at least before pushing :P

kristijanhusak:
This is good idea, but it may be problem when `renderRest` is called, it will find those fields and render them, because they will be excluded and not marked as already `rendered`. Maybe instead of creating exclude property that holds these fields, just to use `remove()` method which will remove those fields from the `fields` array.

theshaunwalker:
I thought about that but it shouldnt be an issue, as it adds the exclude check to the views as well so as long as the field is still in the `$exclude` array then the field shouldnt get rendered in any scenario.

I dont want to remove the fields completely, just exclude them from the general "render all the form" functions as it provides the flexibility to cherry pick fields I'm going to do other stuff with.

My current use-case is I exclude a field from one form's `render()` because I render that specific field elsewhere on the page.

kristijanhusak:
You're right. Merging it. Thanks!

PR #79, opened 2015-04-19
update views to use correct name

### LINKED ISSUE #78: Errors not showing in child form fields
It looks like the real_name is getting set to the individual field's name, when in the HTML the name includes the name of the child form too because its in array format.

Testing out which version of this breaks it as it seems to have broken since I composer updated it from being on an old version.

Update:

Yeah the issue is with
```
$errors->get(array_get($options, 'real_name', $name))
```
which was brought in a commit about "fixing errors in child forms" (ironically lol)

In my case, I have a child form under 'invoice' with a text field 'title'. This is keyed in laravel's error bag as 'invoice.title'. The 'real_name' of the field is 'title' and the `$name` equals 'invoice[title]'.

Is this working for you?

Basically the field needs a dot notation name including the names of its parent form(s) to get the error from `$errors`.

theshaunwalker:
Fixes #78 
Replaced all the 
```
array_get($options, 'real_name', $name)
```
with
```
$name
```
for grabbing errors from Laravels error bag.

Multiple Choice doesnt work with this method though, for a field with a "true" name of `form_name[field_name]` the `$name` in the view becomes `form_name[field_name][]` for HTML multiple choice to work obviously.

So at the moment multiple choice fields wont display error messages.

This is a short fix imo, I feel the design could use having views work off a `$htmlName` or whatever which can be manipulated. So `$name` can be always consistent with domain logic.

P.S. Fields edited but not used personally (so dont know how it effects):
- Collection
- Radio (though simple enough should be fine)

### DIFF
--- a/src/Kris/LaravelFormBuilder/Fields/FormField.php
+++ b/src/Kris/LaravelFormBuilder/Fields/FormField.php
@@ -105,6 +105,7 @@ abstract class FormField
             $this->template,
             [
                 'name' => $this->name,
+                'nameKey' => $this->getNameKey(),
                 'type' => $this->type,
                 'options' => $options,
                 'showLabel' => $showLabel,
@@ -195,6 +196,16 @@ abstract class FormField
         $this->name = $name;
 
         return $this;
+    }
+
+    /**
+     * Get dot notation key for fields
+     *
+     * @return string
+     **/
+    public function getNameKey()
+    {
+        return $this->transformKey($this->name);
     }
 
     /**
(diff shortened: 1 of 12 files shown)


theshaunwalker:
Kept it on the same PR. This fixes the issue with child form errors.

kristijanhusak:
Validating array of data is complicated anyway in Laravel, no matter if this package is used or not.

theshaunwalker:
Yeah this isnt about validating array data, I know that always requires custom stuff for anything but size checks pretty much.

kristijanhusak:
I think maybe at first replacing `[]` with empty string will do the trick for size validation(Haven't tested it though):
```php
public function getDotName()
{
    return str_replace(['[]', '[', ']'], ['', '.', ''], $this->name);
}
```

theshaunwalker:
Realised `FormField` has a `transformKey()` like the Laravel form builder does, so renamed things to `$nameKey` and removed the property.

kristijanhusak:
Great, Thanks!

theshaunwalker:
Thanks for being so receptive to getting things sorted out ^_^

kristijanhusak:
No problem, I want to fix all things that are broken. Thanks for your contribution :)

PR #85, opened 2015-04-23
Rename form:make to make:form for consistency with other Laravel generators

hackel:
This is your call, but I thought I would suggest this since I keep finding myself typing the command backward.


kristijanhusak:
Great idea. I'm not using commands so often, but it should be in that namespace. Thanks.

PR #91, opened 2015-05-02
fire event when rendering

kodok-buncit:
hi kris, i need to modify form before its rendered. i was trying to figure out how to do that without altering your code buat i just dont get it.
so i put this little piece of code on render function.


kristijanhusak:
Hi Kodok-buncit,
I can not accept this PR because tests are failing and facade is used in the code.

You can modify form however you want before you call `form($form)`, because that function calls `renderForm()` in which you put this event fire. I was planning to add events anyway.

How you need to modify it and when? I can help you.

PR #92, opened 2015-05-02
add event before rendering and add new method beforeAdd and afterAdd

kodok-buncit:
add event before rendering and add new method beforeAdd and afterAdd

### DIFF
--- a/src/Kris/LaravelFormBuilder/Form.php
+++ b/src/Kris/LaravelFormBuilder/Form.php
@@ -134,6 +134,78 @@ class Form
     }
 
     /**
+     * add field to the form before specified field
+     *
+     * @param $name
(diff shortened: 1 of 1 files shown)


kristijanhusak:
Same thing as on previous PR, tests are failing because of Events facade.

Just remove it, and clean these functions a bit (Not 100% PSR-2, unused $tmp variables, bad slicing in `addAfter`). 

If you want, i can add these functions myself, or If you want, just clean it up a bit and i'll merge it.

kodok-buncit:
ya please add this function because i need it :+1: 
thanks for your awesome package

kristijanhusak:
@kodok-buncit ok, will do it soon. Closing.

PR #93, opened 2015-05-09
Update Form.php

MeeDNite:
Add Localization for Labels.

### DIFF
--- a/src/Kris/LaravelFormBuilder/Form.php
+++ b/src/Kris/LaravelFormBuilder/Form.php
@@ -691,15 +691,17 @@ class Form
      */
     protected function setupFieldOptions($name, &$options)
     {
+        if (!isset($options['label'])) {
+		    $options['label'] = $this->formHelper->formatLabel($name);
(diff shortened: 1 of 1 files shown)


kristijanhusak:
Good idea, but i cannot accept this PR for several reasons:
1. Tests are failing
2. Facade is used in the code. It should be injected through DI
3. I believe there is no need to check with `has` method, because trans will return it as it is if there is no translation for it.

MeeDNite:
Hi,
Its doesnt use field Name for translation, it translates "label"
and about other issues, I'll be working on it and come back with a better PR.

Thanks.

kristijanhusak:
Yes, but if no label is provided, it will use field name. If you are manually setting label for each field to translate it, it's totally same if you will do `trans('label')` or just `label`. I don't see any use of that.

It's even better to leave it as it is so developer can choose to use it or not. Advantage of that is to avoid unnecessary calls to the Translator class, if no translation is needed for the app.

PR #106, opened 2015-06-06
Update renderUntil() to work with child forms

theshaunwalker:
Hey,

This follows up on past child form issues re: `getName()` getting the html-ified name of `child_form_name[actual_field_name]`

Swapped out `getName()` for `getRealName()` which I added in previous PR so that you're comparing just the `actual_field_name` so it works when doing things like 
```php

	{!! form_until( $form->child_form->getForm() , 'field_name' ) !!}
```

Works as intended now for both forms and child-forms.

P.S.
In the example above you can see its a bit janky passing a child form, because child forms are `ChildFormType` rather than just a `Form` a `getForm()` call is needed to get the underlying form instance. We could add in a check in the relevent helper methods so you can just pass the child form itself and it still work, something like:
```php
if( $form instanceOf ChildFormType ) $form = $form->getForm();
```
Dunno if that creates too much "magic" for your liking though.

### DIFF
--- a/src/Kris/LaravelFormBuilder/Form.php
+++ b/src/Kris/LaravelFormBuilder/Form.php
@@ -273,7 +273,7 @@ class Form
 
         $i = 1;
         foreach ($fields as $key => $value) {
-            if ($value->getName() == $field_name) {
+            if ($value->getRealName() == $field_name) {
                 break;
             }
             $i++;


kristijanhusak:
Hi, thanks for the fix.

Idea for fetching form from chilf form in helper is good, if you want, open another PR with that addition.

PR #107, opened 2015-06-06
Fix default value bug

theshaunwalker:
`! $defaultValue` isnt secure enough and causes type issues for example attempting to set a string of `"0"` evaluates as no default value.

The value of `$defaultValue` when no option is passed, is `false` so just make a hard check for that because nobody should try and set the `default_value` option as "no default value" because that's the default.

### DIFF
--- a/src/Kris/LaravelFormBuilder/Fields/FormField.php
+++ b/src/Kris/LaravelFormBuilder/Fields/FormField.php
@@ -95,7 +95,7 @@ abstract class FormField
             $this->valueClosure = $defaultValue;
         }
 
-        if ((!$defaultValue || $defaultValue instanceof \Closure) && !$isChild) {
+        if (($defaultValue === false || $defaultValue === null || $defaultValue instanceof \Closure) && !$isChild) {
             $this->setValue($this->getModelValueAttribute($this->parent->getModel(), $name));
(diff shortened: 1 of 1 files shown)


theshaunwalker:
```php
$this->assertEquals('lorem', $form->songs->getChild(0)->title->getOption('default_value'));
```
Failed, but from what I can see there shouldnt be a default_value set. So you'll have to take a look when you get the chance.

theshaunwalker:
Just a follow up on this, I'm fiddling about with some stuff re: getting models to work in a child form in a certain situation.
So dont merge this yet as I'll probably have something to replace it.

kristijanhusak:
You are right. Will look into this at tuesday, because i'm not near computer, typing from phone. I'm leaving this now as it is, mostly as remijder, and will probably close it because tests are failing at this point. Thanks for report.

theshaunwalker:
Revisited this and fixed it up.

Solves the issue of a string of `"0"` resolving as `null` thus not reflecting intentions.

kristijanhusak:
Thanks! I appreciate your contribution :-)

theshaunwalker:
Thanks to you @kristijanhusak for your quick replies on a weekend, and making this awesome package to begin with :D

PR #108, opened 2015-06-06
Add form composition

theshaunwalker:
This PR adds the ability to take a child form and add its fields directly to the parent form.

Added `makeField()` as an abstracted method of generating the `FormField`
Added `compose` which is used the same way as `add()` just only needing `$options` and the `$modify` flag. It creates the child form as usual via the new `makeField()` function, then just adds all the fields from that child form, to the parent form.

I've thrown this together as it fixed my issue of trying to use child forms to help with composition of forms but it messed up using model binding, and this is the best solution anyway I just hadnt thought of it till now :P

### DIFF
--- a/src/Kris/LaravelFormBuilder/Form.php
+++ b/src/Kris/LaravelFormBuilder/Form.php
@@ -233,13 +233,16 @@ class Form
             $fields = $class->getForm()->getFields();
         } elseif (is_string($class)) {
             // If its a string of a class make it the usual way
-            $childForm = $this->makeField('', 'form', $options);
-            if (! $childForm instanceof Fields\ChildFormType) {
-                throw new \InvalidArgumentException("Field [{$name}] is not a child form");
+            $options['model'] = $this->model;
+            $options['name'] = $this->name;
+
+            $form = $this->formBuilder->create($class, $options);
+            if (! $form instanceof Form) {
+                throw new \InvalidArgumentException("[{$name}] is not a form");
             }
-            $fields = $childForm->getForm()->getFields();
+            $fields = $form->getFields();
         } else {
-            throw new \InvalidArgumentException("\$class is invalid. Please provide either a string, Form or ChildFormType");
+            throw new \InvalidArgumentException("[{$class}] is invalid. Please provide either a string, Form or ChildFormType");
         }
         foreach ($fields as $field) {
             $this->addField($field);
--- a/src/Kris/LaravelFormBuilder/Form.php
+++ b/src/Kris/LaravelFormBuilder/Form.php
@@ -104,9 +104,28 @@ class Form
 
         return $this;
     }
+    
+    /**
+     * Create the FormField object
+     *
+     * @param string $name
+     * @param string $type
+     * @param array  $options
+     * @return $this
+     */
+    public function makeField($name, $type = 'text', array $options = [])
+    {
+        $this->setupFieldOptions($name, $options);
+
+        $fieldName = $this->getFieldName($name);
+
+        $fieldType = $this->getFieldType($type);
+
+        return new $fieldType($fieldName, $type, $this, $options);
+    }
 
     /**
-     * Add a single field to the form
+     * Create a new field and add it to the form
      *
      * @param string $name
      * @param string $type
@@ -121,24 +140,28 @@ class Form
                 'Please provide valid field name for class ['. get_class($this) .']'
             );
         }
-
-        if (!$modify && !$this->rebuilding) {
-            $this->preventDuplicate($name);
-        }
-
+        
         if ($this->rebuilding && !$this->has($name)) {
             return $this;
         }
 
-        $this->setupFieldOptions($name, $options);
-
-        $fieldName = $this->getFieldName($name);
-
-        $fieldType = $this->getFieldType($type);
-
-        $this->fields[$name] = new $fieldType($fieldName, $type, $this, $options);
-
+        $this->addField($this->makeField($name, $type, $options), $modify);
+        
         return $this;
+    }
+
+    /**
+     * Add a FormField to the form's fields
+     *
+     * @param FormField $field
+     * @return $this
+     */
+    public function addField(FormField $field, $modify = false)
+    {
+        if (!$modify && !$this->rebuilding) {
+            $this->preventDuplicate($field->getRealName());
+        }
+        $this->fields[$field->getRealName()] = $field;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


mikeerickson:
+1

kristijanhusak:
Great idea! Maybe just to set it up to allow first parameter to be the actual class name of the child form, instead of passing it as options, it's more expressive imho **->compose('App/Forms/ChildForm')**. 

I'll not be able to look into this until tuesday, currently typing from phone, so if you want to make this change, and fix some PSR issues in your code, i'll gladly merge it, or i'll set it up compeletely on tuesday. Also will find solution for model binding.

theshaunwalker:
I kept it as the options array as the first solution simply because I know there will still be a need to pass a data array at the very least, I didnt look through the other options to see what else would be made redundant as I know a lot of it relates to HTML.

perhaps `->compose(string $class, array $data, $modify = false)` would be a better footprint, but I think keeping it as the options array covers any future changes/additions which then suddenly makes composition less effective because you cant pass whatever new option has been added.

kristijanhusak:
Yes, that's what i ment, first param class name, second options, and third modify. Then in compose method do proper options merging before making field.

Also, what data structure do you have in the model when you compose forms like this? Are they on the same level as form, or nested?

theshaunwalker:
Do you mean structure of the fields in the final form?
It just appends the child form's fields to to created form's array, so pretty much whatever fields are added in the child's `buildForm()` are essentially treated as if they were called by the parent form calling `compose()`. It doesnt work recursively so if the child itself has child forms of its own, they would remain as the usual nested child forms in the parent form.

Theres too many same words referencing different things here :P

kristijanhusak:
I understand the form structure, but you told in the first comment that form model binding is messed up, so i'm asking are you passing proper data in the model for the composed form fields, or the model still contains data for those fields in the namespaced key as it is in regular child form?

theshaunwalker:
Oh ignore the comment about the model not working, it wasnt working because the model data wasnt namespaced for the child form. My use case that led to this was that fields in the in the parent form and the child form were all for properties of 1 model, wheras a child form expects to have a relation

kristijanhusak:
Is that the desired behavior? I think it will make more sense if it requires structure as you tried at first time. What do you think?

theshaunwalker:
Yes theres not a problem with the form model binding afaik, though I havnt personally tried out model binding that has a relationship that matches with a child form.

The entire issue was I was using child forms incorrectly with regards to model binding, this composition functionality is what I have needed all along, (I've only just recently started using model binding with this package so only just figured it out)

TL:DR;
Everything works as intended, this composition stuff would just be a really nice feature :)

kristijanhusak:
Ok, great :-) fix those things we discussed, and i'll check it out at monday, tuesday and merge it in.

Can you just give me the example usage for this? I mean, what's the biggest value of it? Modularity, separation of concrerns?

Just go mention, this can be also done with class inheritance, where you extend another form class, and call parent::buildForm. It's not the same, but it's one type of solution to similar problem :-)

theshaunwalker:
Its essentially like traits for forms, I have cases where I have the same group of fields across multiple forms (which represent the same the same data structure) so instead of duplicating the code in every form I want to use them in, composition like this is a good solution. So when it comes to updating how I build one of the fields, I just edit the child form instead of all the forms its used in.

Inheritance could be used to do similar things though you're right its not the same, calling composition like in the PR allows to add multiple different forms and be able to specify the order in which they're built.

Do you use PSR-2?

kristijanhusak:
Understood, thanks for explanation :-)

Yes, whole package follows PSR-2, so just set up PR according to that.

theshaunwalker:
No problem, I'll sort the PR out tomorrow, there might be a couple things still wrong after as I dont stick to coding standards myself so might miss some bits T_T lol

kristijanhusak:
Ok. Just a hint, use **$this->formBuilder->create** in the compose method to create form, pass the current model, and merge the fields. That way **makeField** method is not needed, and model binding will work as expected (all data on same level), no need for nesting.

theshaunwalker:
Ah cool, I would still recommend implementing `makeField()` or similar as the as instantiating a new field and adding it to a form are separate concerns, theres no unified way to be able to just build a new form field class as its just new'ed within `add()`

kristijanhusak:
Sure, leave it :-) just maybe move the check for valid name from **add** to **makeField**, i don't like the idea od adding fields with empty name.

theshaunwalker:
Re: the `$name` check

If we pass a `$name` string to `makeField()` it causes all the fields in the child form we're trying to merge, to be prefixed with that name, as it will create the child form as if its going to be used exactly as a child form. Wheras if the `$name` is empty there is simply no prefixing so it works for the composition.

I'm adding another function `addField()` which will take a form field and add it to the array of fields, just further abstraction thats needed to give the `Form` class the flexibility needed.

kristijanhusak:
You do not need child form to do composing, you can create that form as any other form with FormBuilder::create, and just merge it's fields to the current form. ChildFormType does the exact same thing, with some additional things.

theshaunwalker:
I should check my github notifications more often :P spent like 45 minutes debugging a wierd test error just to figure that out :P

theshaunwalker:
All sorted, squashed and cleaned up ready for review :)

kristijanhusak:
Thanks, will check it out from computer tomorrow night when i get home :-)

src/Kris/LaravelFormBuilder/Form.php
```diff
@@ -192,6 +214,38 @@ public function addAfter($name, $fieldName, $type = 'text', $options = [], $modi
...
+        $options['class'] = $class;
+
+        // If we pass a ready made form just extract the fields
+        if ($class instanceof Form) {
+            $fields = $class->getFields();
+        } elseif ($class instanceof Fields\ChildFormType) {
+            $fields = $class->getForm()->getFields();
+        } elseif (is_string($class)) {
+            // If its a string of a class make it the usual way
+            $childForm = $this->makeField('', 'form', $options);
```
kristijanhusak: Instead of making ChildForm, you could create regular form instance here like this:
```php
$options['model'] = $this->model;
$options['name'] = $this->name;

$form = $this->formBuilder->create($class, $options);
$fields = $form->getFields();
```

I don't like the idea of creating fields with empty names, even if only parts are used. This solves the problem.

theshaunwalker: Ah yes I didnt get round to updating that to do that, will ammend.

theshaunwalker:
If this is good to go I'll squash down and push the final version.

I have a PR waiting for after this as I need to get this into master then clean up the new PR git history

kristijanhusak:
Everything good now, merging to master. Thanks!

PR #111, opened 2015-06-09
Proper handling of default values and new explicit value setting

theshaunwalker:
Right sorted after cleaning up some git mess.

TODO:
This breaks the test `it_can_add_child_form_as_field`, related to something with the collection type not child forms as child forms still work as expected.
I've never even looked at the collection stuff before so left it for now as I have zero experience with them.

To review:
`setValue()` I updated the closure to pass the parent's model along with it as it's helpful when dealing with complicated value resolving.
I wanted to double check with you why you pass an empty array to the closure instead of just null? Same for the `Form`'s `$model` default is an empty array. imo it should be null.

#### Changes introduced

Changes behaviour of the setting `'default_value'` and adds the setting `'value'`.
`'default_value'` is now treated as just a fallback value if no value can be found by other means. It also is overridden by a model value in any case, so if the model value is `null` then the field value will be set to `null`. I made this change as I believe if you're passing a model to a form, you're using the form to edit existing data and thus all values would already be set and `'default_value'` should only be used on a blank form for creating new models.

`'value'` setting explicitly sets the value of a field, overriding `'default_value'` AND any attached model.

`'value'` is now the container for whatever value the field needs to be filled with. previously it all just ran off `'default_value'` so all views where applicable have been updated (special note that this will need to be pointed out in changelog as will bork up peoples installations if they've published the views of the package). So anything that modifies / updates a form field value should interact with `'value'` and ignore `'default_value'`

Added `$modelValue` property to `FormField`, just an easy convenient way of storing the model value for that field. Now a field object will have a default value, explicit value and model value available at all times to reference. Part of the trouble before hand was the value logic was pretty much all down to the one little if block, and shoved the resolution back into `'default_value'`. Just adding more flexibility again.

I also added a couple tests to cover this functionality, I havnt added any related to model stuff as the use of global functions borks testing.

I think that pretty much covers it, any questions let me know, and give me some feedback re: the collection stuff seemingly breaking when you get the chance and I'll make whatever changes needed.

### DIFF
--- a/src/Kris/LaravelFormBuilder/Fields/FormField.php
+++ b/src/Kris/LaravelFormBuilder/Fields/FormField.php
@@ -56,11 +56,25 @@ abstract class FormField
     protected $formHelper;
 
     /**
-     * Name of the property for default value
+     * Option name for setting the value
      *
      * @var string
      */
-    protected $valueProperty = 'default_value';
+    protected $valueProperty = 'value';
+
+    /**
+     * Option name for the default value
+     *
+     * @var string
+     */
+    protected $defaultValueProperty = 'default_value';
+
+    /**
+     * Value set on accompanying model
+     *
+     * @var mixed
+     */
+    protected $modelValue = null;
 
     /**
      * Is default value set?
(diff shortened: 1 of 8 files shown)


theshaunwalker:
Oh I've just remembered I cut out the `$isChild` checks. I'll add them back in xD

kristijanhusak:
I tried to figure out why tests are failing, and it's probably because other types have different names for default value, and they are using it. You tried to rebuild value setting, but it depends on the type. Renaming `default_value` to `value`, and adding `default_value` in the `render` method just as check should be enough. I'll try to do it myself to see how it works.

theshaunwalker:
Yes tests pass for me too.

This does revert making the model value override the default value even if the model value is null.

Do you agree that if a model is passed to a form default values should be ignored?

My argument is that if you're passing a model you're using the form to update existing values, whereas default values should only be used on a blank form when creating a new model.

kristijanhusak:
@theshaunwalker 
This can be closed, since it's done in https://github.com/kristijanhusak/laravel-form-builder/pull/112 .

Many thanks for help, I really appreciate it :)

PR #112, opened 2015-06-10
Rename default_value to value, and use default_value as fallback value.

kristijanhusak:
This is another version of this PR https://github.com/kristijanhusak/laravel-form-builder/pull/111.

@theshaunwalker 
I done this other way, and all tests are passing. Also added those 2 tests from your PR, and they are passing properly.
There were more places to change this, that's why your PR's tests are failing.

I would really appreciate if you could test this PR by yourself to see if it's working.

### DIFF
--- a/src/Kris/LaravelFormBuilder/Fields/CheckableType.php
+++ b/src/Kris/LaravelFormBuilder/Fields/CheckableType.php
@@ -18,7 +18,7 @@ class CheckableType extends FormField
     {
         return [
             'attr' => ['class' => null, 'id' => $this->getName()],
-            'default_value' => null,
+            'value' => null,
             'checked' => false
         ];
     }
--- a/src/Kris/LaravelFormBuilder/Fields/ChildFormType.php
+++ b/src/Kris/LaravelFormBuilder/Fields/ChildFormType.php
@@ -33,7 +33,7 @@ class ChildFormType extends ParentType
     {
         return [
             'class' => null,
-            'default_value' => false,
+            'value' => false,
             'formOptions' => [],
             'data' => [],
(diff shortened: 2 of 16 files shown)


theshaunwalker:
Yes tests pass for me too. Sorry I overlooked this I havnt used the different field types that much.

This does revert making the model value override the default value when model value is null.

Do you agree that if a model is passed to a form default values should be ignored?

My argument is that if you're passing a model you're using the form to update existing values, whereas default values should only be used on a blank form when creating a new model.

kristijanhusak:
I'm not sure what's the best option.
I think it makes sense like it is now, because it will be used only if none of the other sources are provided and valid (value property and model value).
I don't believe it will be used much, can not even think of one use case at this moment, probably setting some default value for select, that can be overridden by a model.

theshaunwalker:
Fair enough, all the use cases that apply to me at the moment during development can be avoided by other means.

This seems good to go then.

theshaunwalker:
Seeing as this is still open I'll mention it here, I think for the value closures it would be better to just add the form itself as the 2nd parameter instead of just the model, I have scenarios where I want to access data on the form.

kristijanhusak:
I don't think any other parameters are needed except value, because you can access form within close with `$this`. Example:
```php
class PostForm extends Form
{
    public function getText()
    {
        return 'some text';
    }

    public function buildForm()
    {
        $this->add('name', 'text', [
            'value' => function($value) {
                return $value . $this->getText();
            }
    }
}
```

theshaunwalker:
Ah! I didnt know that, the closure change can be reverted then.

PR #119, opened 2015-06-18
Update composer.json

gildniy:
I think it'll be better if instead of using illuminate/html we use laravelcollective/html as this later is getting support and update, what is not the case for the first!

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,7 +10,7 @@
     ],
     "require": {
         "php": ">=5.4.0",
-        "illuminate/html": "5.0.*@dev",
+        "laravelcollective/html": "5.1.*",
         "illuminate/database": "5.*@dev"
     },
     "require-dev": {


kristijanhusak:
What's the advantage of this one against Laravel's except the support? I couldn't notice any other difference.

gildniy:
I was just suggesting to adopt that official replacement, so that the theme package could not do that duplication, bcz in case someone is using the package from laravel collective, this wll one day make conflict in Facades!

I like your package here BTW!

kristijanhusak:
I'll test it out now to see if everything is working properly with it. If yes, will replace it.

Closing this PR, since changing composer.json is not enough, ServiceProvider also needs update, and this is set to allow only laravel 5.1.*, i need to support 5.0 also. Thanks for letting me know about this package.

PR #123, opened 2015-06-26
Update README.md

Gounlaf:
* Fix wrong namespace for `SongsController.php`
* Fix blade example, replacing `@extend` by `@extends` (and link to the default `app` layout)
* Add information about routes


kristijanhusak:
Thanks!

PR #134, opened 2015-07-23
Tweak label formatting for greater flexibility

FourStacks:
Hi @kristijanhusak 

Just started formatting all the forms on an app where I'm using the form builder and just noticed (for the first time!) that all labels are converted to title case which isn't really the most flexible output IMO.  Would be better to output as sentence case and if the user want to they can use CSS to convert to title case if required.  Switching to sentence case is likely to output a closer representation of what the author has written for the label in most cases

### DIFF
--- a/src/Kris/LaravelFormBuilder/FormHelper.php
+++ b/src/Kris/LaravelFormBuilder/FormHelper.php
@@ -239,6 +239,6 @@ class FormHelper
             return null;
         }
 
-        return ucwords(str_replace('_', ' ', $name));
+        return ucfirst(str_replace('_', ' ', $name));
     }
 }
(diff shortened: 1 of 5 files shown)


kristijanhusak:
Hi @FourStacks ,

Yeah, i guess you are right. Not sure how i determined to put it as ucwords. Only problem is that it may cause really tiny minor BC break, but i believe most of people set custom labels. Can you please just fix all failed tests? Thanks!

FourStacks:
Yeah, I did think about the BC implications but I figured it was very minor and that (as you say) most people will have set custom labels (though perhaps not so much for choice field labels).  In those cases though I'd imagine most people would have masked the output to change to sentence case with CSS anyway so in that case it wouldn't make much difference.  Sorry about the tests btw - was being lazy and making the change on github.  Have done things properly now so CI should come up ok with the latest commit

kristijanhusak:
Great! Thanks!

Issue #135, opened 2015-07-28
Form handling/validating


barryvdh:
I was looking at this package as a replacement for Symfony Forms in Laravel.
I kind of like Symfony Forms, but it's a pretty much tied to Twig + Symfony conventions.

Something I do like in Symfony is the handling of forms (http://symfony.com/doc/current/book/forms.html#handling-form-submissions)

```php
// just setup a fresh $task object (remove the dummy data)
$task = new Task();

$form = $this->createFormBuilder($task)
    ->add('task', 'text')
    ->add('dueDate', 'date')
    ->add('save', 'submit', array('label' => 'Create Task'))
    ->getForm();

$form->handleRequest($request);

if ($form->isValid()) {
    // perform some action, such as saving the task to the database

    return $this->redirectToRoute('task_success');
}
```

So this takes the POST data, calls the setters on the Task instance and checks the constraints etc.

In Laravel, I would think of something like this:

```php
$task = new Task();

$form = \FormBuilder::plain([
        'model' => $task,
    ])
    ->add('task', 'text', [
        'rules' => 'required|min:10'
    ]);

$form->handle();

if ($form->isValid()) {
    $task->save();

    return redirect()->back();
}
```

This would have a few benefits (above the FormRequests)

- Rules coupled to the Form (so you don't miss a form or add a rule for something that doesn't exist or has a different name)
- Could possibly use the Label from the field. (https://github.com/kristijanhusak/laravel-form-builder/issues/56)
- Might be safer, because you only set the fields you have defined in your form (vs `Input::all()` or fillable/guarded)

And would open the door for something I really like from Former (https://github.com/formers/former); Client-side HTML5 validation. Eg. a `required` rule -> `required` attribute, `max` rule -> `max-length` attribute etc. This would be automatic, so no need to require twice.

What do you think about this? I would be open for helping with this, when you think it's a good idea (and I'm going too switch to this)

kristijanhusak:
Hi @barryvdh ,

That is a great feature, and I also like it. I wanted to implement it at some point, but binding the values from the form to the model is a bit tricky because of the nested child forms/collection types. I just didn't have enough time to find best solution to that.
I will probably have some time this weekend to look into it.

barryvdh:
Okay cool, let me know if you need any help!

About child forms: They would get their own Model/data object, so could be handled similarly by the ChildForm handle method?

[...]

PR #140, opened 2015-08-09
HTML5 validation attributes

barryvdh:
As discussed in #135 it's now possible to add validation rules and validate forms.

This PR automatically transforms rules to attributes. So a `required` rule will add a `required` attribute, `max` -> `maxlength` or `max`, depending on the type, etc

For pattern validation, the validation message is used to set the `title` attribute, so the browser can show a better message.

Validation types currently supported:
 - [x] alpha
 - [x] alpha_dash
 - [x] alpha_num
 - [x] integer
 - [x] numeric
 - [x] required
 - [x] min
 - [x] max
 - [x] between
 - [x] in
 - [x] not_in
 - [x] image
 - [x] accepted 
 - [x] boolean
 - [x] after, before (date handling)
 - [x] digits, digits_between
 - [x] size
 - [x] regex (But not sure about the correctness)
 - [x] mimes (Not actual mime types, but extensions)

Validations that could perhaps be added, but I skipped because of complex patterns.
- date, date_format (date formats)
- ip address (pattern validation?)

See https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/HTML5/Constraint_validation#Validation-related_attributes

Not sure about the cleanest implementation or that this should be configurable? If we make sure the HTML5 attributes match the Laravel rules, it shouldn't affect existing users, right?

### DIFF
--- a/src/Kris/LaravelFormBuilder/Fields/FormField.php
+++ b/src/Kris/LaravelFormBuilder/Fields/FormField.php
@@ -4,6 +4,7 @@ use Illuminate\Database\Eloquent\Collection;
 use Illuminate\Database\Eloquent\Model;
 use Kris\LaravelFormBuilder\Form;
 use Kris\LaravelFormBuilder\FormHelper;
+use Kris\LaravelFormBuilder\RulesParser;
 
 /**
  * Class FormField
@@ -232,6 +233,11 @@ abstract class FormField
             $options['label_attr']['class'] .= ' ' . $this->formHelper
                 ->getConfig('defaults.required_class', 'required');
             $options['attr']['required'] = 'required';
+        }
+        
+        if ($rules = $this->getOption('rules')) {
+            $rulesParser = new RulesParser($this);
+            $options['attr'] += $rulesParser->parse($rules);
         }
 
         $options['wrapperAttrs'] = $helper->prepareAttributes($options['wrapper']);
--- a/src/Kris/LaravelFormBuilder/RulesParser.php
+++ b/src/Kris/LaravelFormBuilder/RulesParser.php
@@ -1,0 +1,601 @@
+<?php namespace Kris\LaravelFormBuilder;
+
+use Illuminate\Support\Arr;
+use Illuminate\Support\Str;
+use Kris\LaravelFormBuilder\Fields\FormField;
+
+/**
+ * Laravel Validator rules to HTML5 attributes parser.
+ *
+ * Based on Laravel Validator and Former LiveValidation
+ * @see https://github.com/laravel/framework
+ * @see https://github.com/formers/former
+ */
+class RulesParser
+{
+    /**
+     * @var FormField
+     */
+    protected $field;
+
+    /**
+     * @var FormHelper
+     */
+    protected $formHelper;
+
+    /**
+     * @param FormField $field
+     */
+    public function __construct(FormField $field)
+    {
+        $this->field = $field;
+        $this->formHelper = $this->field->getParent()->getFormHelper();
+    }
+
+    /**
+     * Parse a rule for an input into an array of attributes.
+     *
+     * @param  string|array $rules
+     * @return array
+     */
+    public function parse($rules)
+    {
+        $attributes = array();
+        $rules = $rule = (is_string($rules)) ? explode('|', $rules) : $rules;
+
+        foreach ($rules as $rule) {
+            list($rule, $parameters) = $this->parseRule($rule);
+
+            if ($rule && method_exists($this, $rule)) {
+                $attributes += $this->$rule($parameters);
+            }
+        }
+
+        return $attributes;
+    }
+
+    /**
+     * Check that a checkbox is accepted. Needs yes, on, 1, or true as value.
+     *
+     *   accepted  -> required="required"
+     *
+     * @return array
+     *
+     * @see http://laravel.com/docs/5.1/validation#rule-accepted
+     */
+    protected function accepted()
+    {
+        return [
+            'required' => 'required',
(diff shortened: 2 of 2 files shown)


barryvdh:
I've added some more validations.

You can test with this:

```php
 public function buildForm()
    {
        $this
            ->add('alpha', 'text', [
                'rules' => 'alpha',
            ])
            ->add('alphanum', 'text', [
                'rules' => 'alpha_num',
            ])
            ->add('alphadash', 'text', [
                'rules' => 'alpha_dash',
            ])
            ->add('numeric', 'text', [
                'rules' => 'numeric',
            ])
            ->add('numericNumber', 'number', [
                'rules' => 'numeric',
            ])
            ->add('integer', 'text', [
                'rules' => 'integer',
            ])
            ->add('integerNumber', 'number', [
                'rules' => 'integer'
            ])
            ->add('min', 'text', [
                'rules' => 'min:3',
            ])
            ->add('minNumber', 'number', [
                'rules' => 'min:3',
            ])
            ->add('max', 'text', [
                'rules' => 'max:5',
            ])
            ->add('maxNumber', 'number', [
                'rules' => 'max:5',
            ])
            ->add('between', 'text', [
                'rules' => 'between:3,8',
            ])
            ->add('betweenNumber', 'number', [
                'rules' => 'between:3,8',
            ])
            ->add('digits', 'text', [
                'rules' => 'digits:3',
            ])
            ->add('digits_between', 'text', [
                'rules' => 'digits_between:3,5',
            ])
            ->add('size', 'text', [
                'rules' => 'size:3',
            ])
            ->add('sizeNumber', 'number', [
                'rules' => 'size:3',
            ])
            ->add('in', 'text', [
                'rules' => 'in:foo,bar',
            ])
            ->add('not_in', 'text', [
                'rules' => 'not_in:foo,bar',
            ])
            ->add('after', 'date', [
                'rules' => 'after:09-08-2015',
            ])
            ->add('before', 'date', [
                'rules' => 'before:09-08-2015',
            ])
            ->add('regex', 'text', [
                'rules' => 'regex:/^\d+$/',
            ])
            ->add('image', 'file', [
                'rules' => 'mimes:jpg',
            ])
            ->add('boolean', 'text', [
                'rules' => 'boolean',
            ])
            ->add('accepted', 'checkbox', [
                'rules' => 'accepted',
            ])
            ->add('required', 'text', [
                'rules' => 'required',
            ])
            ->add('submit', 'submit');

    }
```

kristijanhusak:
Wow, nice! Really clean and understandable code. I haven't tested it yet, but will do that tomorrow.
I'm rarely using html5 validation, since it's not fully supported by all browsers, especially those regex patterns and stuff. Do you use some additional JS libraries to make this work properly in all browsers?

About the configuration, I would like to allow users to disable adding those html5 attributes, just with a simple flag on Form class, like we have for showing errors under fields (check [here](https://github.com/kristijanhusak/laravel-form-builder/blob/master/src/Kris/LaravelFormBuilder/Form.php#L52))
And also maybe to set it to be off by default, and set it on in some near future. Don't want people to suddenly get bunch of html5 attributes in their markup. Or even if we set it on by default, just to be possible to disable it. Validation is really "young" in here (since few days ago), so setting it even on by default should be ok.

barryvdh:
Okay, I've added the last couple of validations (mime and digits for numbers) and tweaked/cleaned it up a bit. Squashed it, so should be good for testing :)

Making it an option would be good :) I'll leave the defaults up to you.

About validation, all the rules work in the latest Chrome. Haven't tested in other browsers, but I guess most of it works in latest IE, Firefox and Safari. It's all based on pure html5 spec, so it doesn't break on older browsers, it just doesn't work.
But there are polyfills that mimic the html5 validation. I usually use https://github.com/jzaefferer/jquery-validation, which has support for (most of) the html5 attributes, so it's a drop in replacement.

Something like this just adds it to all forms and uses the bootstrap styling to show messages:

```js
     $('form').each(function() {
        $(this).validate({
            errorClass: "has-error help-block",
            validClass: "", //has-success to make it green when okay
            errorElement: "span",
            highlight: function(element, errorClass, validClass) {
                $(element).closest('.form-group').removeClass(validClass).addClass(errorClass).find('span.' + errorClass).addClass('help-block');
            },
            unhighlight: function(element, errorClass, validClass) {
                $(element).closest('.form-group').removeClass(errorClass).addClass(validClass);
            }
        });
    });
```

kristijanhusak:
Great! Will you add some unit tests for the RulesParser class?

Also, can you add that flag for disable/enable, and set it to true(enabled) by default?

After the merge, i'll add that example to the docs, so people can set it up quick and easy.

Thank you!

kristijanhusak:
I tested it, it works great!
Also tried that polyfill, and noticed that pattern validation is not working correctly. Saw that you opened issue/PR to fix it. 

This can be merged without any problem.

If you are not able to add disable/enable option and tests, I will handle that.
If you are, just let me know when it's finished so i can check it out and merge it.
Thanks!

barryvdh:
Hmm, that PR was merged quite some time ago, so if it doesn't work, I might have to check again. 

I'm not sure when I have time to fix some tests, perhaps tonight but pretty busy. If you could add the options, that's probably faster (not sure where to put it etc).

For tests, I'll have to see if I can make some time. But I added the expected behaviour in the phpdocs  already.

kristijanhusak:
Sure, i'll add the options.

About the tests, I checked the code once more, and I think there is no need to write them for now, code is pretty simple and straightforward, constructing arrays from rules. I will add that in some near future.

Merging this in. I really appreciate your help and thank you once more!

PR #141, opened 2015-08-09
Process closure default values correctly

theshaunwalker:
Started getting a wierd bug where closure default values in options were throwing errors on render because escaping strings was just interpreting them as objects because they were unresolved closures.

The root was some missing logic when the closures were actually processed, basically ignoring default values at that point.

This fixes it anyway.

### DIFF
--- a/src/Kris/LaravelFormBuilder/Fields/FormField.php
+++ b/src/Kris/LaravelFormBuilder/Fields/FormField.php
@@ -100,7 +100,13 @@ abstract class FormField
     protected function setupValue()
     {
         $value = $this->getOption($this->valueProperty);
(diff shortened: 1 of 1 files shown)


theshaunwalker:
Thought there was a better way to do this but from diving in the code this is the better way, test needs fixed but dont have the time to do it right now, this was a quick fix I've just pushed up so you can see it.

kristijanhusak:
@theshaunwalker 
closures for default value are not supported, i thought there is no need for closures in that situation. Can you give me an example where they can be useful?

theshaunwalker:
Now that you mention it @kristijanhusak I remember that use-case being discussed. I'll close this, I just had existing code that seemed to stop working but I think it was from changes on my own branch.

PR #144, opened 2015-08-13
Remove regex rule

barryvdh:
I know I added this myself, but I don't feel really confident about this. Regexes are difficult and I don't think we can simply transform php regex to html patterns. It might give okay results for easy cases, but unpredictable for other cases. Best to just let the developer set his own pattern if he needs this.


kristijanhusak:
Yeah, i guess you are right. Some things are not supported in html pattern (one of lookaheads or lookbehinds).

PR #147, opened 2015-08-19
Make it possible to resolve the entity option with the IoC

CaporalDead:
With Symfony2 you can do this :

```php

use Doctrine\ORM\EntityRepository;
// ...

$builder->add('users', 'entity', array(
    'class' => 'AcmeHelloBundle:User',
    'query_builder' => function(EntityRepository $er) {
        return $er->createQueryBuilder('u')
            ->orderBy('u.username', 'ASC');
    },
));

```

With Eloquent, the concept of repositories does not exists, so with your package we can do this :

```php

$this->add('languages', 'entity', [
        'class' => 'App\Language',
        'property' => 'short_name',
        'query_builder' => function (App\Language $lang) {
            // If query builder option is not provided, all data is fetched
            return $lang->where('active', 1);
        }
    ]);

```

If we look into your ```EntityType``` class, we can see that you new up an instance of the "class" without any parameter and it's perfectly fine, but if we use your package in a bigger project we probably use the repository pattern (as I do in my current project).

In most cases, a repository will take the ```App``` instance or the model that we'll use in the repository itself. A little tweak to make it possible in your package could be to use the ```app()``` method (from the helpers functions) to get the object that we want.

Example : 

```php

$this->add('languages', 'entity', [
        'class' => 'App\Repository\LanguagesRepository',
        'property' => 'short_name',
        'query_builder' => function (App\Repository\LanguagesRepository $repo) {
            return $repo->getActiveLanguages();
        }
    ]);

```

Not sure if it's clear enough but I'll be glad to discuss about it with you :).

### DIFF
--- a/src/Kris/LaravelFormBuilder/Fields/EntityType.php
+++ b/src/Kris/LaravelFormBuilder/Fields/EntityType.php
@@ -50,7 +50,7 @@ class EntityType extends ChoiceType
                 $data = $data->lists($value, $key);
             }
         } else {
-            $data = [];
+            $data = $class->lists($value, $key);;
         }
 
         if ($data instanceof Collection) {
--- a/src/Kris/LaravelFormBuilder/Fields/EntityType.php
+++ b/src/Kris/LaravelFormBuilder/Fields/EntityType.php
@@ -29,12 +29,12 @@ class EntityType extends ChoiceType
             return parent::createChildren();
         }
 
-        $entity = $this->getOption('class');
+        $class = $this->getOption('class');
         $queryBuilder = $this->getOption('query_builder');
         $key = $this->getOption('property_key');
         $value = $this->getOption('property');
 
-        if (!$entity || !class_exists($entity)) {
+        if (!$class || !class_exists($class)) {
             throw new \InvalidArgumentException(sprintf(
                 'Please provide valid "class" option for entity field [%s] in form class [%s]',
                 $this->getRealName(),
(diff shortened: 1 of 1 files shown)


CaporalDead:
Didn't test pass the unit test, I'll look @ it tonight BTW.

kristijanhusak:
Hi @CaporalDead ,
Thanks for the PR.

I understand your requirements, but I don't like the idea of tight coupling framework functionality with the package (`app()` function).

Symfony's repositories are there by default, and in Laravel they are optional. If you will use repository to get the the values for the dropdown (or whatever else), you can just pull it in through the constructor, and set it to ChoiceType:

```php

class MyForm extends Form
{
    /**
     * @var LanguageRepository
     */
    protected $repository;

    public function __construct(LanguagesRepository $repository)
    {
        $this->repository = $repository;
    }

    public function buildForm()
    {
        $this->add('languages', 'choice', [
            'choices' => $this->repository->getActiveLanguages()
        ]);
    }
}
```

Just make sure to return array from the repository method, and everything will work out just fine.

CaporalDead:
I understand why you don't want to use the ```app()``` function but your package is already tightly coupled with the Laravel framework. Instead of using the ```app``` function maybe we could inject the ```Illuminate\Contracts\Container\Container``` into the form field via the ```FormField``` class which will implements an interface like ```ContainerAwareInterface``` and use a trait ?

kristijanhusak:
It is only in views (Laravel Form facade), which i also plan to change. Everything else is done through dependency injection.

What is the value of this PR? I mean, in which way is your solution better than example I posted in last comment? I really need good arguments to approve something like this, because it's really specific, and most of the times not needed.

PR #161, opened 2015-09-28
`renderUntil` method - throw exception if passed a non-existent field name

pnoonan:
I spent a couple hours debugging a problem which was caused by a typo in a field name passed to the "form_until" helper function, which just wraps "renderUntil". Everything after that point was basically getting rendered twice. Ideally such a case would just be caught immediately and an exception would be thrown. Tests are passing for me.

### DIFF
--- a/src/Kris/LaravelFormBuilder/Form.php
+++ b/src/Kris/LaravelFormBuilder/Form.php
@@ -292,7 +292,7 @@ class Form
             return $this;
         }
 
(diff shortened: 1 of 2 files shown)


kristijanhusak:
Yeah, i didn't write test for that, missed it. Thanks for the fix!

pnoonan:
No problem, thanks for the merge!

PR #164, opened 2015-10-05
Update form-class-template.stub namespace position

koenvu:
Changes the position of the namespace definition to be the same as the new default Laravel style.


kristijanhusak:
Thanks!

PR #166, opened 2015-10-06
Add template prefix functionality

koenvu:
This pull request adds functionality to define a template prefix for a form. It should be fully backwards compatible.

The goal of this change is the ability to quickly copy an entire views/form directory and create an entirely new styling for a form. Using this new style for all the form fields in a form is as simple as setting the `template_prefix` option for the `Form`. This template prefix trickles down into the `getViewTemplate` method for `FormField` as well.

3 tests have been added to check functionality.

If you have any questions or think something should work/be configured differently, I'm happy to discuss!

### DIFF
--- a/src/Kris/LaravelFormBuilder/Fields/FormField.php
+++ b/src/Kris/LaravelFormBuilder/Fields/FormField.php
@@ -126,7 +126,7 @@ abstract class FormField
      */
     protected function getViewTemplate()
     {
-        return $this->getOption('template', $this->template);
+        return $this->parent->getTemplatePrefix() . $this->getOption('template', $this->template);
     }
 
     /**
--- a/src/Kris/LaravelFormBuilder/Form.php
+++ b/src/Kris/LaravelFormBuilder/Form.php
@@ -737,6 +737,16 @@ class Form
     }
 
     /**
+     * Get template prefix that is prepended to all template paths
+     *
+     * @return string
+     */
+    public function getTemplatePrefix()
+    {
+        return $this->getFormOption('template_prefix', $this->formHelper->getConfig('template_prefix'));
+    }
+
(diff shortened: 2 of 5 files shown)


kristijanhusak:
Hi @koenvu ,

Thanks for the pull request. Everything looks ok, but i think this functionality is not really necessary since prefix can be easily changed from the configuration file when you publish the package. Just change all `laravel-form-builder::` with whatever you want and you are ready to go (check [here](https://github.com/kristijanhusak/laravel-form-builder/blob/master/src/config/config.php#L14). Am I missing something?

koenvu:
This change makes it easy to have two different sets of form stylings in two different locations. So say I have an administrator section with a different layout from the main site I can easily make two sets of views and switch between them by setting a `template_prefix` option.

koenvu:
So, to be more specific: one `div`-based form and one `table`-based form for example.

kristijanhusak:
Yeah, that makes sense. But check needs to be added, to see if prefix is defined, because if you set prefix, and do not change template names, than you can get something like `test::laravel-form-builder::text`.

koenvu:
Would that be a check for `laravel-form-builder` specifically or any package prefix at all? Because for the latter I might see it break code that would be perfectly valid. For the first it seems like a thing we could do, even if it is a little odd to make a specific check for the prefix (despite the fact it might not change in the future).

But I'm open to either! What would have your preference?

kristijanhusak:
You are right, it's really hard to check. Let's leave it like this than. Merging it in. Thanks once more.

koenvu:
Just noticed that the `template_prefix` option spills over to the form tag as an HTML attribute. Will fix that and submit a new pull request tonight ;-)

PR #167, opened 2015-10-07
Cleanup template_prefix code

koenvu:
Before using setFormOptions it would not correctly pull template_prefix
from the options. Leaving it in caused it to be displayed as an
attribute on the form.

I made a separate setter for the template prefix and followed the logic
used for other settings in setFormOptions.

Sorry for the inconvenience and the oversight in my previous pull
request!


kristijanhusak:
Great, thanks!

PR #172, opened 2015-10-12
Create NewAPI.php

imanghafoori1:
I know is not something perfect or may look somewhat hacky code.
Any way, I am eager to know what you think about it.
Let's move on. :+1:

### DIFF
--- a/src/Kris/LaravelFormBuilder/NewAPI.php
+++ b/src/Kris/LaravelFormBuilder/NewAPI.php
@@ -7,6 +7,8 @@
     
     
     /**
+     * 
+     * 
      * @method FieldOptionSetters text( string $field_name, array $options = [], boolean $modify = false )
      * @method FieldOptionSetters textarea( string $field_name, array $options = [], boolean $modify = false )
      * @method FieldOptionSetters radio( string $field_name, array $options = [], boolean $modify = false )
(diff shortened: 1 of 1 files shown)


src/Kris/LaravelFormBuilder/NewAPI.php
```diff
@@ -0,0 +1,125 @@
...
+	 * @method FieldOptionSetters time(      string  $field_name, array $options = [] , boolean $modify = false)
+	 * @method SelectFieldOptionSetters select_(string  $field_name, array $options = [] , boolean $modify = false)
+	 */
+	class NewAPI extends Form
+	{
+		public $fieldName;
+		public function __call($fieldType,$args = [])
+		{
+			$permitted = [
+			    'text','textarea','radio','checkbox','select',
```
kristijanhusak: Why is this missing `form`, `collection`, and several other fields? Those could be also added.
Also, there is no need to hardcode these here, you can get all available from [here](https://github.com/kristijanhusak/laravel-form-builder/blob/master/src/Kris/LaravelFormBuilder/FormHelper.php#L37)

kristijanhusak:
Can you please update the code to be according to PSR-2 standard?

Also, i left one comment for now in the code, so check it out.

kristijanhusak:
@imanghafoori1 
Thanks for the PR, but I think this is not really necessary in the core. You can easily include this file as helper in your project.

PR #187, opened 2015-12-08
fixed falsy checkboxes not being saved by adding a hidden input with the same name

koichirose:
I probably did not find out how to handle this because it should be a pretty common issue: unchecked checkboxes won't be included in the POST data, so after checking a checkbox, it's impossible to uncheck it.

One can check each field for existence in the controller, or simply add a hidden input with the same name BEFORE the checkbox as I did here.

### DIFF
--- a/src/views/checkbox.php
+++ b/src/views/checkbox.php
@@ -5,6 +5,7 @@
 <?php endif; ?>
 
 <?php if ($showField): ?>
+    <?= Form::hidden($name, false) ?>
     <?= Form::checkbox($name, $options['value'], $options['checked'], $options['attr']) ?>
 
     <?php include 'help_block.php' ?>


kristijanhusak:
Yes, this is a common issue, but i cannot just merge this in for two reasons:

1. It's a minor BC break - some users are probably handling this in some way and not expect this to appear
2. hidden value is hardcoded as false, but someone may want to use something else.

I would suggest you to:

* Publish the views and add this line manually, or
* Manage missing values in service where you do the model saving (controller, form handler, whatever) - This is a bit more complicated if you have bunch of checkboxes, but one or two can be easily handled with `$request->get('checkbox', 'default value here')`.

koichirose:
Got it. I'm using this on a completely new project so I'm not worrying about BC.

Thanks

PR #193, opened 2015-12-21
Update doc for CollectionType

CaporalDead:
Add doc for "empty_row" option on CollectionType
Update "options" default value according to code


kristijanhusak:
Thanks!

PR #194, opened 2015-12-27
little fixes for laravelcollective/html 5.2

SpadXIII:
Found 2 api-changes in laravelcollective/html that stopped artisan from working. Missing was `$app['view']`.


kristijanhusak:
Thanks!

PR #200, opened 2016-01-04
Rudimentary fix to avoid exceptions in Laravel 5.1

marcoraddatz:
... through LaravelCollective\HTML in 1.6.40.

Don't know if you already implemented a version check.


kristijanhusak:
I haven't added a version check, thought it is not really needed. Thanks.

Issue #149, opened 2015-08-24
Migrated to new validation system; old tests don't pass because of empty request


marcoraddatz:
I migrated all validation rules to your new system, which I like a lot. However, my old tests won't pass, because I'm told, the required fields haven't been filled out – but they are.

Here's a small example:

AddressForm (I have two more fields and a submit button in this class, but have cut them for this example):
```php
$this
    ->add('address_1', 'text', [
            'required'  => true,
            'label'     => trans('label.address_1'),
            'rules'     => 'required|max:255',
        ]
    )
    ->add('address_2', 'text', [
            'required'  => false,
            'label'     => trans('label.address_2'),
            'rules'     => 'max:255',
        ]
    )
    ->add('city', 'text', [
            'required'  => true,
            'label'     => trans('label.city'),
            'rules'     => 'required|max:255',
        ]
    )
    ->add('zip_code', 'text', [
            'required'  => true,
            'label'     => trans('label.zip_code'),
            'rules'     => 'required'
        ]
    );
```

AddressController → update:
```php
if (!$validation->isValid()) {
    print_r($request->all());
    print_r($validation->getErrors());
    die();
    return $this->redirectWithErrors($validation);
}

[...]
```

Test:
```php
$address = factory(App\Address::class)->make()->toArray();
$address['address_1'] = 'User 1 Address 1 - Updated';

$this
    ->actingAs(User::find(3))
    ->visit('/address/1/edit')
    ->see(trans('address.edit.title'))
    ->see('User 1 Address 1')
    ->submitForm(trans('label.submit'), $address) // Here's the problem
    ->see(trans('address.edit.success'))
    ->see('User 1 Address 1 - Updated')
;
```

Test output:
```
Array
(
    [_method] => PUT
    [_token] => qxad1oeCsKvILU5vFzGNgJ5JkNC6hMh93MtiJ1BV
    [address_1] => User 1 Address 1 - Updated
    [address_2] => Apt. 602
    [city] => Marienberg
    [zip_code] => 82756
    [country_id] => 276
)
Array
(
    [address_1] => Array
        (
            [0] => The Address Field 1 field is required.
        )

    [city] => Array
        (
            [0] => The City field is required.
        )

    [zip_code] => Array
        (
            [0] => The ZIP code field is required.
        )

    [country_id] => Array
        (
            [0] => The Country field is required.
        )

)
```

[...]

PR #201, opened 2016-01-04
Avoid exceptions on non-existing fields

marcoraddatz:
I also removed the test.

This refers to #149.


kristijanhusak:
I totally forgot about it. Thanks!

Issue #206, opened 2016-01-12
Automatic translation of labels


koenvu:
I'd like to propose (and have a discussion about) functionality for allowing automatic translation of labels. Currently (by default) the label is determined by taking the name, converting underscores to spaces and transforming the first letter to uppercase.

## My proposal:

- add a new option to `Form`: (something similar to) `language_name`
- if this option is set change the `FormHelper::formatLabel` method to (something similar to) `trans(language_name . '.' . $name)`
  - if the option is not set, the current behavior will be used

## Example

Imagine a file `resources/lang/en/contact_form.php`:

```
<?php

return [
    'first_name' => 'Your first name'
];
```

A form:

```
$plain = FormBuilder::plain(['language_name' => 'contact_form']);
$plain->add('first_name', 'text');
```

The label of the field would then be **Your first name**.

## Questions

@kristijanhusak (and others) is this something you'd want? If so, any input on the names of the options? Let me know, I'd love to create a pull request for this feature!

kristijanhusak:
@koenvu 
I think this is a useful feature, but i avoided it for only one reason - one more dependency to inject.

I'ts not really good to just hardcode laravel `trans()` function into the package. It needs to be injected.

I would suggest injecting it into FormHelper class, and then using it where needed.

Feel free to create PR, and we will discuss on it.

koenvu:
I was planning on injecting indeed, the `trans` was just a (bad) example :+1:

mohsensaeedi:
I used this way:
`
use Lang;
`

then use Lang::get as below
`
->add('username', 'text', [
                'label' => Lang::get('messages.username'),
                'rules' => 'required',
])
`

PR #208, opened 2016-01-12
Add localizable names/label functionality

koenvu:
As discussed in kristijanhusak/laravel-form-builder#206

I'm not sure this is the best way to go about this, but I'm submitting this pull request to start a discussion.

I've opted not to do the generation of the localizable name in the `FormHelper`, because there is no real connection to the form. In addition to that, `FormHelper` is defined as a singleton so setting a language string would make little sense in there.

I have also removed a piece of code from the `choices` field. As described in that commit:

> The array passed is a key value array. To me, it doesn't make sense to pass this value through the label formatter, since it will change (intended) underscores to spaces.

Furthermore, I have only 'enabled' implicit localization for labels that are implicitly determined. I.e.: for select and choice labels, they are passed directly using their key value arrays.

### DIFF
--- a/src/Kris/LaravelFormBuilder/FormHelper.php
+++ b/src/Kris/LaravelFormBuilder/FormHelper.php
@@ -1,9 +1,10 @@
 <?php  namespace Kris\LaravelFormBuilder;
 
-use Illuminate\Contracts\View\Factory as View;
-use Illuminate\Database\Eloquent\Model;
 use Illuminate\Support\Collection;
+use Illuminate\Translation\Translator;
+use Illuminate\Database\Eloquent\Model;
 use Kris\LaravelFormBuilder\Fields\FormField;
+use Illuminate\Contracts\View\Factory as View;
 
 class FormHelper
 {
@@ -12,6 +13,11 @@ class FormHelper
      * @var View
      */
     protected $view;
+
+    /**
+     * @var Translator
+     */
+    protected $translator;
 
     /**
      * @var array
@@ -72,9 +78,10 @@ class FormHelper
      * @param View    $view
      * @param array   $config
      */
-    public function __construct(View $view, array $config = [])
+    public function __construct(View $view, array $config = [], Translator $translator)
     {
         $this->view = $view;
+        $this->translator = $translator;
         $this->config = $config;
         $this->loadCustomTypes();
     }
@@ -222,6 +229,10 @@ class FormHelper
     {
         if (!$name) {
             return null;
+        }
+
+        if ($this->translator->has($name)) {
+            return $this->translator->get($name);
         }
 
         return ucfirst(str_replace('_', ' ', $name));
--- a/src/Kris/LaravelFormBuilder/FormHelper.php
+++ b/src/Kris/LaravelFormBuilder/FormHelper.php
@@ -76,9 +76,10 @@ class FormHelper
 
     /**
      * @param View    $view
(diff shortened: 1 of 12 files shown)


src/Kris/LaravelFormBuilder/FormHelper.php
```diff
@@ -72,9 +78,10 @@ class FormHelper
      * @param View    $view
      * @param array   $config
      */
-    public function __construct(View $view, array $config = [])
+    public function __construct(View $view, array $config = [], Translator $translator)
```
kristijanhusak: Please move `$translator` to be 2nd param, and `$config` to be last, because it's optional.

kristijanhusak:
Thanks for the PR. It looks ok. We didn't discuss what should happen with the child forms? How will child form inherit language_name? 

For example, if we have `PostForm`, and we have `CommentForm` as a child form, where `language_name` is not provided for the child form, should `language_name` be inherited, for example like `post_form`?
And if we provide language_name for child form (eg. `comment_form`), should it be `comment_form` or `post_form.comment_form` ? 

I'm not really sure what's the best solution, but my opinion is to have it inherit the `language_name` from parent form, and if it is provided for child form, use that. If someone needs nesting like that, they can provide it in the child form options like `'language_name' => 'post_form.comment_form'`. What are your thoughts?

koenvu:
I agree, letting it bubble over from parent to child form unless the child form explicitly sets the `language_name`. I will update this pull request to reflect this.

koenvu:
On second thought, it looks like this can be a bit tricky, since the label is determined upon creation of the field (by ways of `allDefaults()` -> `'label' => $this->formHelper->formatLabel($this->parent->getLocalizableName($this->getRealName()))`). So when the child form is created the labels are set, and not altered when the child form is then added to the 'parent' form.

What would work best for you? Not inheriting the language in the child form, or somehow postponing setting the label (for example, determining it when preparing for rendering)?

kristijanhusak:
Just pass it to child form like everything else is done (example [here](https://github.com/kristijanhusak/laravel-form-builder/blob/master/src/Kris/LaravelFormBuilder/Fields/ChildFormType.php#L87)). Just make sure to cover both string and instance.

kristijanhusak:
@koenvu 
This looks good for now. I just found one issue with this. On several places you are calling this:
```php
$this->formHelper->formatLabel($this->parent->getLocalizableName($this->getRealName()))
```
The problem is that `formatLabel` is called every time, but we need it only when there is no `language_name`. For example, imagine u set a language name to `post_form`, and your field name is `first_name`. Generated label for translation would be `post_form.first_name`, which is ok. But when we run that through `formatLabel`, we get `post form.first name`, which is wrong.

It would be good to move `formatLabel` call into the `getLocalizableName`, and call it only if there is no `language_name` provided for the form.

kristijanhusak:
@koenvu If you do not have time to finish this, i can merge this and finish what is left. Just let me know.

koenvu:
@kristijanhusak Sorry, spare time is limited this week. I can pick it up beginning of next week, or you can merge and finish it up. Either is fine with me :+1:

kristijanhusak:
@koenvu Ok, i'll probably merge it today and fix what's left. Thanks for letting me know.

afflicto:
I really need this for a multilingual project. Any thoughts when will this be available in stable?

PR #209, opened 2016-01-12
Add functionality to add required class based on rules

koenvu:
As discussed in kristijanhusak/laravel-form-builder#206

This change should be fairly straightforward and shouldn't break existing code. The only side effect might be that the `required` class is added where the user does have the `required` rule but didn't have the `required` option before.

### DIFF
--- a/src/Kris/LaravelFormBuilder/Fields/FormField.php
+++ b/src/Kris/LaravelFormBuilder/Fields/FormField.php
@@ -216,6 +216,8 @@ abstract class FormField
         $helper = $this->formHelper;
         $rulesParser = new RulesParser($this);
         $rules = $this->getOption('rules');
+        $parsedRules = $rules ? $rulesParser->parse($rules) : [];
+
         $this->options = $helper->mergeOptions($this->options, $options);
 
         if ($this->getOption('attr.multiple') && !$this->getOption('tmp.multipleBracesSet')) {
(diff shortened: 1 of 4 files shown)


src/Kris/LaravelFormBuilder/Fields/FormField.php
```diff
@@ -223,7 +227,7 @@ protected function prepareOptions(array $options = [])
             $this->addErrorClass();
         }
 
-        if ($this->getOption('required') === true) {
+        if ($this->getOption('required') === true || $rulesParser->containsRequired($rules)) {
```
kristijanhusak: There is no need for special method on rules parser to check for that, you can only parse the rules, and do `in_array('required', $parsedRules)`.

koenvu: I did it this way because the `parse` method returns attributes, and I wasn't sure if that wouldn't be susceptible to change in the future. However, if you decide you do indeed want to use `parse` I'll change the code :-)

kristijanhusak: There shouldn't be any problem with this in the future. So please change it as i explained.
Just use `isset($rules['required'])` instead of `in_array()` or `array_search()`, key is much more important.

kristijanhusak:
It looks good. Thanks!

PR #216, opened 2016-02-05
Add check for nested translations

paultela:
Laravel translations can be nested and thus the translator will return
an array for these types of translations.  Passing an array into the
FormBuilder where is it expecting a string results in an error when the
form is rendered due to htmlentities not knowing what to do with an
array.  This change akes sure that the returned translation is actually
a string before using it.

I also updated the test to check for this case.

### DIFF
--- a/src/Kris/LaravelFormBuilder/FormHelper.php
+++ b/src/Kris/LaravelFormBuilder/FormHelper.php
@@ -233,7 +233,11 @@ class FormHelper
         }
 
         if ($this->translator->has($name)) {
-            return $this->translator->get($name);
+            $translatedName = $this->translator->get($name);
+
+            if (is_string($translatedName)) {
+                return $translatedName;
+            }
         }
 
         return ucfirst(str_replace('_', ' ', $name));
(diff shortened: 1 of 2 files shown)


kristijanhusak:
Thanks!

paultela:
Just curious if a new release with this fix is planned soon?  I got bit by this on another project today, the work around isn't difficult but would be nice to have it working nonetheless.

kristijanhusak:
I totally forgot this is not tagged. It is the only thing that is fixed after the latest release. I will try to fix several more things before releasing a new version. You can use dev-master for now.

marcoraddatz:
Thanks for the feature.

I didn't see that the Readme got updated. Is this sth. we should do? As I understand the changes, FormBuilder now tries to guess the name of a form based on existing translations, right?! Or how do you want it to work?

kristijanhusak:
@paultela This is published in latest version. Thanks for the PR.

@marcoraddatz this is just a precaution, since translations can be nested. If you want to nest your form translations you can use `language_name` option on form. It's not documented, but i will add something to the docs soon. You can get more information on this issue https://github.com/kristijanhusak/laravel-form-builder/issues/206

kristijanhusak:
@marcoraddatz here's some documentation for that http://kristijanhusak.github.io/laravel-form-builder/form/basics.html#form-options

marcoraddatz:
This is great. Thank you!

PR #220, opened 2016-02-24
Add default classes to a field.

jvdlaar:
It's currently only possible to set one class for a form field. When you set a default class in your laravel-form-builder.php file and then set a new class in your form builder array then only the last class will be added.

In this commit I've created the following things:

1. Default classes on the field wrapper for all fields (form-TYPE) in addition to the class that you set in wrapper_class.
2. Additional class on the field wrapper for choice fields (form-TYPE and form-type-CHOICETYPE
3. Wrapper class on the choice options wrapper (form-CHOICETYPE)
4. You can add additional classes without overwriting the defaults (I've added a array_merge_map_recursive function for this).


jvdlaar:
Before I spent time fixing the errors, can you tell me if this is something that might get included?

kristijanhusak:
Hi,

Thanks for the PR, and sorry for delay. I like the idea, but implementation could be better. This works only for wrapper, and i think we should avoid custom array merge recursive functions and try to use default ones with different implementation. Try to refactor it, and think about adding these defaults to other things, not just wrapper.

jvdlaar:
Sorry for the late reply, the code works for all classes. Not only the wrapper class. So this also works:

    $form->add('name', 'text', [
      'label' => trans('organization.name'),
      'rules' => ['required', 'min:2', 'max:255'],
      'attr' => ['class' => 'bla'],
      'label_attr' => ['class' => 'bla'],
    ]);

It just adds some default classes for the wrapper class.
I will make an attempt at removing the custom array_merge.

jvdlaar:
I have updated my patch. I've created default classes for labels and fields and adapted the config.php for this. I've left them empty by default because the classes on the wrapper should be sufficient most of the time.

I've removed the array_merge_map_recursive and put all code in mergeOptions. I don't think it's possible to do this with a default php array function. The problem is that `array_replace_recursive` replaces all strings and that this: `array_merge_recursive(array('class' => 'foo'), array('class' => 'bar'))` becomes `array('class' => array('foo', 'bar'))`. So something needs to turn it into a string again.

The tests are all fixed again.

Can you take a look at it and tell me if you have a different idea for mergeOptions?

jvdlaar:
@kristijanhusak can you have a look at my latest attempt?

src/Kris/LaravelFormBuilder/Fields/FormField.php
```diff
@@ -493,9 +493,28 @@ protected function setDefaultOptions(array $options = [])
...
+    /**
+     * Creates default wrapper classes for the form element.
+     *
+     * @param array $options
+     * @return array
+     */
+    protected function setDefaultClasses(array $options = [])
+    {
+        return [
+            'wrapper' => ['class' => $this->formHelper->getConfig('defaults.' . $this->type . '.wrapper_class', 'form-' . $this->type)],
```
kristijanhusak: why `getConfig` here has default value as `'form-' . $this-type ` ?

kristijanhusak:
Hey @jvdlaar , sorry for delay, i didn't had time to check it out properly. I left you one comment on the code. It looks ok, i'll just test it out manually and it should be fine.

jvdlaar:
It seemed to me a sensible default as the wrapper now gets a class so all child items in css can inherit that class. But I guess this change can break backwards compatibility, so perhaps it should be empty to not break existing installations.

kristijanhusak:
Yeah BC is my concern. Please set it up to default to empty string so we do not break BC.

jvdlaar:
Will do, thanks for the review!

jvdlaar:
I've tried to fix some things such that they are more consistent with how things worked before.

jvdlaar:
Did you perhaps have time to look into this?

kristijanhusak:
Hey @jvdlaar , sorry, i was at vacation for 2 weeks without any internet connection.
This looks great, i'm just worried about BC breaks, since this change appends to default classes, instead of replacing it. I'll keep this as PR for another week or so until i fix other things, so i will publish a minor release. Thank you very much!

rudiedirkx:
Maybe I'm too late, but it would be nice if the 'option merger' would be extensible. If another strangely mergable option comes along, it would only require a simple method override, instead of a huge merge method. Maybe a method `mergeOption($name, $first, $second)` ? Or maybe next time =)

jvdlaar:
No need to applogize for going on vacation ;-)

@rudiedirkx I will create a second PR after this one is in to make it extensible. I myself was thinking about default validations.

kristijanhusak:
Merging this, since i'll be publishing new version soon. Thank you @jvdlaar !

kristijanhusak:
@jvdlaar just got this issue https://github.com/kristijanhusak/laravel-form-builder/issues/271.
Isn't this suppose to be overridden if provided? 
No matter if you define a default per type or not, it will always append provided class to the default one.

```php
// config.php
return [
'defaults'      => [
        'wrapper_class'       => 'form-group',
        'wrapper_error_class' => 'has-error',
        'label_class'         => 'control-label',
        'field_class'         => 'form-control',
        'help_block_class'    => 'help-block',
        'error_class'         => 'text-danger',
        'required_class'      => 'required'
        // Override a class from a field.
        'text'  => [
            'wrapper_class'   => 'form-field-text',
            'label_class'     => 'form-field-text-label',
            'field_class'     => 'form-field-text-field',
        ]
    ],
]
```

```php
// Field should get class "form-field-text-field"
$this->add('name', 'text');
```

```php
// Field should get class "form-control"
$this->add('name', 'textarea');
```

```php
// Field should have only "my-class"
$this->add('name', 'text', [
  'attr' => ['class' => 'my-class']
]);
```

```php
// Field should have only "my-textarea-class"
$this->add('name', 'textarea', [
  'attr' => ['class' => 'my-textarea-class']
]);
```

It breaks BC pretty bad, and there was PR https://github.com/kristijanhusak/laravel-form-builder/pull/257 that added ability to append to a default class, so this is not a desired behavior.
The problem is [here](https://github.com/kristijanhusak/laravel-form-builder/blob/master/src/Kris/LaravelFormBuilder/FormHelper.php#L136) i believe, which concats the classes, no matter what you do. I think we should go back to array_replace_recursive, since we do not want classes to be concated, only used if not overriden.

jvdlaar:
Yes that is caused by the PR. It appends all classes that are specified "upstream". Perhaps for BC we should make a flag that enables / disables this behavior? So by default it replaces classes, if the flag is enabled then it appends classes?

Because quite often it's quite useful to have the classes appended. For example you can quite easily make this: All form elements have form-group class, all textareas and text fields have form-group text-input and all radios / checkboxes have form-group choice-input. And one textfield has form-group text-input autocomplete.

But I agree that BC is broken now. But it would be very cool that you can still achieve above result.

kristijanhusak:
we can leave the "default class by type" thing, it's great feature, but if you pass class when creating a field with `add()`, than i want that value to have bigger priority over the configuration. 
So if you set for textarea in config to have `form-group textarea`, and you create field with just  `add('desc', 'textarea')`, than it will use that, but if it is overridden in options, do not append it to the defaults, completely overwrite it.
If you want to append to a certain field some class, without typing the whole thing, use `class_append` which i linked in previous comment.
Makes sense?

jvdlaar:
Yeah that sounds like a decent solution.

kristijanhusak:
Ok great, thanks. I'll fix it, just wanted to check with you if that was initially desired behavior.

vovech:
Hi People, just a question this has been solved in some tag? I've installed 1.8 and can't change default class it always adding form-control class for checkbox. Thanks in advance.

Modify defaults: 

`'defaults' => [
	 ...
	'checkbox'  => [
		'field_class'     => '', // null or false same behavior
	]        
]`

`$this->add('name', 'checkbox', [
	'label'   => 'test', 
	'attr' => [
		'class'   => false
	],
]);`

Still get form-control class:

`<input class="form-control" id="" name="name" type="checkbox">`

kristijanhusak:
@vovech why are you passing `'class' => false`, if you already disabled it in the config?
What happens if you remove that part, and just rely on the config part?

vovech:
@kristijanhusak it always get form-control class. I tried with 1.9 version en it's working perfectly as expected. Seems than is problem of 1.8 only. By now i'm ok with 1.9 so thank you :)

PR #236, opened 2016-06-20
Prepared FormBuilder to be compatible with Laravel 5.3+

marcoraddatz:
When I fixed FormBuilder to be compatible with Laravel < 5.2, I implemented a stupid fix that might throw an exception on Laravel 5.3+.


kristijanhusak:
Thanks!

Issue #246, opened 2016-07-14
Add Form->getValues()


rudiedirkx:
I'd like my Form to know it's incoming values. Not the entire request, but only the known and validated values. Would make a very useful addition in postback. Instead of the very unsafe `$request->all()`, you'd get the safe data from `$form->getValues()` after validating.

My custom code has a hack, but it's enormous because it can't be integrated into your `Form`, where it belongs.

rudiedirkx:
PR: https://github.com/kristijanhusak/laravel-form-builder/pull/248

rudiedirkx:
Done! Thanks Kristijan!

PR #248, opened 2016-07-30
#246 +Form->getValidatedValues()

rudiedirkx:
Adds a public `$form->getValidatedValues()` function to extract all Request values according to the submitted form, not according to the submitted data. Uses all fields' `getAllAttributes()` to collect a flat list of all attributes, including all child forms and theirs etc.

The main function (`getValidatedValues`) is somewhat of a misnomer, because it doesn't actually validate, and the returned values don't have to be validated in order to call it, but I couldn't find a better name.

**Usage:** instead of

    $values = $request->all();

call

    $values = $form->getValidatedValues();

### DIFF
--- a/src/Kris/LaravelFormBuilder/Fields/ChildFormType.php
+++ b/src/Kris/LaravelFormBuilder/Fields/ChildFormType.php
@@ -41,6 +41,15 @@ class ChildFormType extends ParentType
     }
 
     /**
+     * @inheritdoc
+     */
+    public function getAllAttributes()
+    {
+        // Collect all children's attributes.
+        return $this->parent->getFormHelper()->mergeAttributes($this->children);
+    }
+
+    /**
      * @return mixed|void
      */
     protected function createChildren()
--- a/src/Kris/LaravelFormBuilder/Fields/FormField.php
+++ b/src/Kris/LaravelFormBuilder/Fields/FormField.php
@@ -587,6 +587,16 @@ abstract class FormField
     }
 
     /**
+     * Get this field's attributes, probably just one.
+     *
+     * @return array
+     */
+    public function getAllAttributes()
+    {
+        return [$this->getNameKey()];
+    }
+
+    /**
      * Get value property
      *
      * @param mixed|null $default
--- a/src/Kris/LaravelFormBuilder/Fields/StaticType.php
+++ b/src/Kris/LaravelFormBuilder/Fields/StaticType.php
@@ -35,4 +35,13 @@ class StaticType extends FormField
             'attr' => ['class' => 'form-control-static', 'id' => $this->getName()]
         ];
     }
+
+    /**
+     * @inheritdoc
+     */
+    public function getAllAttributes()
+    {
+        // No input allowed for Static fields.
+        return [];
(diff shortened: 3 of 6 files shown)


kristijanhusak:
This looks good. Can you just please add at least 1 test that checks this feature. Thanks.

rudiedirkx:
Oooh tests. Yeah, I'll give that a go =)

rudiedirkx:
Added 1 test method that tests `Form::getAllAttributes()` and `Form::getValidatedValues()`.

koenvu:
Perhaps `$form->getSubmittedValues()` instead of `$form->getValidatedValues()` would make more sense? Thoughts?

rudiedirkx:
I agree 'valdated' is a misnomer, but 'submitted' is too, because the point is to not return all submitted values, but only the known form elements values, and NILs for those missing. `getSubmittedValues` is fine by me, but it's not perfect either =)

koenvu:
@rudiedirkx True, but that's why we ask the form for submitted values, not the request ;-). Either way it is tough naming this one... Perhaps a third option would be `getProcessedValues`... but it does not really process anything. `$form->getFormValues()` has the word form in it two times, which looks strange too.

How about a simple `$form->getValues()` perhaps, or `$form->getContainedValues()`?

rudiedirkx:
Haha, I'll leave this one up to Kris. I'm good with all your ideas too. I just want to see the functionality in. If you see anything off in code itself, this is the time.

rudiedirkx:
If we're changing the name, we should probably change the test too: `it_returns_validated_values`

koenvu:
Could probably get away with `it_returns_only_request_values_for_fields_within_the_form` in that case 👍

kristijanhusak:
I have another suggestion :D - `$form->getFieldValues()`, 
but also partially agree with @koenvu on `$form->getFormValues()`.

`getValidatedValues` is wrong because we should get values for all fields, not just the ones that are validated.

`getSubmittedValues` can be misunderstood because people can expect it to return all submitted values, even if they are not part of the form.

koenvu:
Oh, I really like the `getFieldValues()` !!

rudiedirkx:
My branch name took all these into account ;)

I'll change it into `getFieldValues`. Any more necessary fixes I can include, or is this it? How is the test?

rudiedirkx:
Ah crap. I called it `getFormValues` now... My own app created `getFieldValues` so I can get behind that too.

@kristijanhusak Let me know if `getFormValues` is good enough =) and other issues with this PR.

kristijanhusak:
It's good, thanks!

koenvu:
As long as we are editing, is it okay to change it to `getFieldValues`? I really like that so much more than `getFormValues` 😄

kristijanhusak:
Oh yes, sorry, @rudiedirkx , set it to `getFieldValues` in the meantime :D

rudiedirkx:
Done and done!

rudiedirkx:
Ah, now it's done. #toomanybranches

kristijanhusak:
Great, thanks!

Issue #243, opened 2016-07-13
Let form definition choose whether to show the label


rudiedirkx:
Currently, the renderer decides whether to show the label, which is `true` by default:

````
public function render(array $options = [], $showLabel = true, $showField = true, $showError = true) {
  ...
  return $this->formHelper->getView()->make(
    $this->getViewTemplate(),
    [
      'name' => $this->name,
      'nameKey' => $this->getNameKey(),
      'type' => $this->type,
      'options' => $this->options,
      'showLabel' => $showLabel,
      'showField' => $showField,
      'showError' => $showError
    ]
  )->render();
````

but the renderer might be called via via via (child form of a child form), which results in the default `true`.

I would very much like it if the form definition could choose whether to show the label:

````
$this->add('organization_address', 'form', [
  'class' => AddressForm::class,
  'label' => 'ORGANIZATION ADDRESS',
  'showLabel' => false,
]);
````

Currently I've solved that by overriding `ChildFormType::render()` and adding

````
$showLabel = $this->getOption('showLabel', $showLabel);
````

but that might be a nice addition for all fields, probably in `FormField::render()`. I see some more `$showLabel` logic there, so it might not be as easy as I think.

kristijanhusak:
I'm currently short on time, so please create a PR with these additions and i'll review it.

rudiedirkx:
Relevant PR: https://github.com/kristijanhusak/laravel-form-builder/pull/249

rudiedirkx:
So no.

PR #249, opened 2016-07-30
#243 +showLabel field option

rudiedirkx:
Very simply takes the optional option from the field config.

This way isn't perfect IMO, because this way the field config overrides the runtime `render()` param. I'd like it to be the opposite: the renderer has final say whether to show the label, and if it doesn't have a preference (NULL), the field config is used. To do that, the `render()` param requires a default NULL value that will be converted to a bool in combination with the option:

```
public function render(array $options = [], $showLabel = null, $showField = true, $showError = true) {
    // ..
    if ($showLabel === null) {
        $showLabel = $this->getOption('showLabel', true); // TRUE is still the default default
    }
```

but that requires the NULL param to be the default everywhere, and that seems impossible, because this method is called from many places, where TRUE is still the default.

So this PR is second best: if the field config exists, it always wins.

### DIFF
--- a/src/Kris/LaravelFormBuilder/Fields/FormField.php
+++ b/src/Kris/LaravelFormBuilder/Fields/FormField.php
@@ -144,6 +144,9 @@ abstract class FormField
         $value = $this->getValue();
         $defaultValue = $this->getDefaultValue();
 
+        // Let the field definition choose whether to show the label.
+        $showLabel = $this->getOption('showLabel', $showLabel);
+
         if ($showField) {
             $this->rendered = true;
         }


kristijanhusak:
There's an option `label_show` that I added not so long time ago https://github.com/kristijanhusak/laravel-form-builder/commit/127d467c92f7dbb7caf58ed919e2e1d389598acd , is that what you wanted to use instead of `showLabel` for `getOption`?

rudiedirkx:
Eeeeeeh, maybe yeah... My approach was to alter `$showLabel`, but you added a new variable in the templates. I've overridden all templates, so I don't see that. I guess this PR is obsolete then. You told me to create a PR!

The test seems wrong though:

    $this->assertNotContains('label', $view);

Shouldn't you check for `Name`? Or is that the `label` inside `<label for=..>`?

kristijanhusak:
Yeah just updated the views. Sorry for asking for PR, i was really short on time when i checked your issue.

For the test, i'm just checking if printed view has any text with label, which would be `<label for=...>` in case the `label_show` is not working.

So can we close this ?

rudiedirkx:
Yup. Closed.

Issue #250, opened 2016-07-30
Doubly nested form field name is wrong


rudiedirkx:
It's not very common, or even sensible, but it should work nonetheless. Let's say you have a form CreateOrganization with a field `organization[entity]` that loads a child form Organization with a field `details[name]`. The resulting HTML element should have a name `organization[entity][details][name]`, because it's doubly double nested. Instead its name attribute is `organization[entity][details[name]]`. The brackets are wrong, because of the doubly double nesting.

With a name like that, PHP is confused and values end up everywhere. The value should be in `$_POST[organization][entity][details][name]`.

The name is already wrong in the FormField constructor, so it happens early on, but I can't find it.

Maybe a solution is to convert all names into dot notation during Form building, and only back to bracket notation when rendering HTML. That might not be very bc though...

It'd also be very nice to have a method in FormHelper to transform from dot notation to bracket notation. There is a `transformToDotSyntax()` but not the opposite.

Like I said, it's a stupid edge case, so no rush.

rudiedirkx:
Relevant PR: https://github.com/kristijanhusak/laravel-form-builder/pull/251

rudiedirkx:
Done!

PR #251, opened 2016-07-30
#250 bugfix in doubly nested field name

rudiedirkx:
Added `FormHelper->transformToBracketSyntax()`, which helps in fixing the field's name. It's not super pretty, but I think it handles all cases.

### DIFF
--- a/src/Kris/LaravelFormBuilder/Form.php
+++ b/src/Kris/LaravelFormBuilder/Form.php
@@ -997,12 +997,17 @@ class Form
      */
     protected function getFieldName($name)
     {
-        if ($this->getName() !== null) {
-            return $this->formHelper->transformToBracketSyntax(
-                $this->formHelper->transformToDotSyntax(
-                    $this->getName() . '[' . $name . ']'
-                )
-            );
+        $formName = $this->getName();
+        if ($formName !== null) {
+            if (strpos($formName, '[') !== false || strpos($formName, ']') !== false) {
+                return $this->formHelper->transformToBracketSyntax(
+                    $this->formHelper->transformToDotSyntax(
+                        $formName . '[' . $name . ']'
+                    )
+                );
+            }
+
+            return $formName . '[' . $name . ']';
         }
 
         return $name;
--- a/src/Kris/LaravelFormBuilder/Form.php
+++ b/src/Kris/LaravelFormBuilder/Form.php
@@ -997,7 +997,11 @@ class Form
     protected function getFieldName($name)
     {
         if ($this->getName() !== null) {
-            return $this->getName().'['.$name.']';
+            return $this->formHelper->transformToBracketSyntax(
+                $this->formHelper->transformToDotSyntax(
+                    $this->getName() . '[' . $name . ']'
+                )
+            );
         }
 
         return $name;
--- a/src/Kris/LaravelFormBuilder/FormHelper.php
+++ b/src/Kris/LaravelFormBuilder/FormHelper.php
@@ -285,6 +285,21 @@ class FormHelper
     }
 
     /**
+     * @param string $string
+     * @return string
+     */
+    public function transformToBracketSyntax($string)
+    {
+        $name = explode('.', $string);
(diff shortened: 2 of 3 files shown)


kristijanhusak:
@rudiedirkx Thanks for the pr. can you add some tests for this? So we can be sure it works properly.

rudiedirkx:
What would the test be? Check output name attributes? It's not really a testable feature. Give me a direction and I'll cook something up.

kristijanhusak:
Setup that form structure in a test (Form has child form, which has another child form inside), and add checks similar to these [here](https://github.com/kristijanhusak/laravel-form-builder/blob/master/tests/FormTest.php#L635)

rudiedirkx:
Oooh, not even with rendering into HTML? I can do that. Probably next week.

rudiedirkx:
Or just now. Voila.

It's kinda annoying you can't run a single class: `vendor/bin/phpunit tests/FormTest.php`, because FormTest assumes the other test has run, which defines the form classes. But it's super super fast, so doesn't really matter.

How's this? Looks a bit skinny, but tests the fix exactly. Test result before fix:

```
There was 1 failure:

1) FormTest::it_has_html_valid_element_names
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'child[form][name][text]'
+'child[form][name[text]]'
```

kristijanhusak:
I just noticed here and another PR that you are using field names like `name[text]`. Is that the only situation where you get names like this `child[form][name[text]]`, or it happens also if you add child form, that has another child form?

I'm asking this because package never supposed to manage that kind of naming. If you need that kind of naming, use a child form.

rudiedirkx:
Child forms alone work, but a child form with nesting doesn't.

I sometimes (in my app, not really a package) use nesting like `name[text]` to group content. For instance a wysiwyg will have a `text` and `format`: `description[text]` and `description[format]`, but it's not really form worthy. Or I can have a form with dynamic inputs, like a translation for every question in the database. The form will dynamically create elements `question[<id>]`, so `question[1]`, `question[34]` etc.

All subsequent validations work fine with these doubly/mixed nesting, it was just the field name that didn't.

kristijanhusak:
Naming of dynamically created elements should be working fine if you are using CollectionType.
Anyway, this is a good addition, but i would like to skip this double conversion if it's not really needed.
Just add a check if `$name` contains `[` or `]`, and reformat it to the proper syntax, if not, just proceed how it was.

rudiedirkx:
I don't think that's more readable, and it won't really matter performance wise...

That would make:

```
if (strstr($this->getName(), '[') || strstr($this->getName(), ']')) {
	return $this->formHelper->transformToBracketSyntax(
		$this->formHelper->transformToDotSyntax(
			$this->getName() . '[' . $name . ']'
		)
	);
}

return $this->getName() . '[' . $name . ']';
```

I really don't like that just to save 2 ms per page... If you do, I'll fix it. Or if you know a nicer way to check if the name contains `[` or `]`.

rudiedirkx:
I didn't know about 'collection'. That might suit better than typing `question[N]`. I'm assuming it accepts rules.

kristijanhusak:
I agree that is not more readable, but why to do transformation it if it's not really necessary. 
I would suggest using `strpos` instead of `strstr`, as php.net says [here](http://php.net/manual/en/function.strstr.php):

> Note:
> If you only want to determine if a particular needle occurs within haystack, use the faster and less memory intensive function strpos() instead.

so something like this should work:

```php
$fieldName = $this->getName();
if (strpos($fieldName, '[') !== false || strpos($fieldName, ']') !== false) {
    return $this->formHelper->transformToBracketSyntax(
        $this->formHelper->transformToDotSyntax(
            $fieldName . '[' . $name . ']'
        )
    );
}

return fieldName . '[' . $name . ']';
```

rudiedirkx:
Voila. Tests still good.

kristijanhusak:
Nice, thanks!

Issue #253, opened 2016-07-31
Add form events


rudiedirkx:
As stated in #252, maybe events is a way to go. Before/after validation seems obvious, but also very useless: when a field has been added, and when a form has been built.

rudiedirkx:
Relevant PR: https://github.com/kristijanhusak/laravel-form-builder/pull/254

rudiedirkx:
It took a few tries, but done!

PR #254, opened 2016-07-31
#253 Added events

rudiedirkx:
Events added:

* AfterFieldCreation (Form, FormField)
* AfterFormCreation (Form)
* BeforeFormValidation (Form, Validator)

I think the names are obvious. With these events, an app or package could do something like:

```
public function onCreateFormField(AfterFieldCreation $event) {
  if ($label = $event->getField()->getOption('label')) {
    $event->getField()->setOption('label', 'x ' . $label);
  }
}

public function onCreateForm(AfterFormCreation $event) {
  $event->getForm()->setFormOption('novalidate', '');
}

public function onValidateForm(BeforeFormValidation $event) {
  $validator = $event->getValidator();
  $validator->after(function() use ($validator) {
    rand(0, 1) and $validator->messages()->add('field', 'Error!');
  });
}
```

### DIFF
--- a/src/Kris/LaravelFormBuilder/Events/BeforeFormValidation.php
+++ b/src/Kris/LaravelFormBuilder/Events/BeforeFormValidation.php
@@ -1,0 +1,40 @@
+<?php namespace Kris\LaravelFormBuilder\Events;
+
+use Illuminate\Contracts\Validation\Validator;
+use Kris\LaravelFormBuilder\Form;
+
+class BeforeFormValidation {
+
+    /**
+     * @var $form Form
+     */
+    protected $form;
+
+    /**
+     * @var $validator Validator
+     */
+    protected $validator;
+
+    /**
+     * Create a new event instance.
+     */
+    public function __construct(Form $form, Validator $validator) {
+        $this->form = $form;
+        $this->validator = $validator;
+    }
+
+    /**
+     * Return the event's form.
+     */
+    public function getForm() {
+        return $this->form;
+    }
+
+    /**
+     * Return the event's validator.
+     */
+    public function getValidator() {
+        return $this->validator;
+    }
+
+}
--- a/src/Kris/LaravelFormBuilder/Form.php
+++ b/src/Kris/LaravelFormBuilder/Form.php
@@ -6,6 +6,7 @@ use Illuminate\Contracts\Validation\Validator;
 use Illuminate\Http\Exception\HttpResponseException;
 use Illuminate\Http\Request;
 use Kris\LaravelFormBuilder\Events\AfterFieldCreation;
+use Kris\LaravelFormBuilder\Events\AfterFormValidation;
 use Kris\LaravelFormBuilder\Events\BeforeFormValidation;
 use Kris\LaravelFormBuilder\Fields\FormField;
 
@@ -1109,7 +1110,11 @@ class Form
             $this->validate();
         }
 
-        return !$this->validator->fails();
+        $isValid = !$this->validator->fails();
+
+        $this->eventDispatcher->fire(new AfterFormValidation($this, $this->validator, $isValid));
+
+        return $isValid;
     }
 
     /**
--- a/src/Kris/LaravelFormBuilder/Form.php
+++ b/src/Kris/LaravelFormBuilder/Form.php
@@ -1,6 +1,6 @@
 <?php namespace Kris\LaravelFormBuilder;
 
-use Illuminate\Contracts\Events\Dispatcher;
+use Illuminate\Contracts\Events\Dispatcher as EventDispatcher;
 use Illuminate\Contracts\Validation\Factory as ValidatorFactory;
 use Illuminate\Contracts\Validation\Validator;
 use Illuminate\Http\Request;
@@ -26,7 +26,7 @@ class Form
     protected $model = [];
 
     /**
-     * @var Dispatcher
+     * @var EventDispatcher
      */
     protected $eventDispatcher;
 
@@ -632,10 +632,10 @@ class Form
     /**
      * Set the Event Dispatcher to fire Laravel events
      *
-     * @param Dispatcher $eventDispatcher
+     * @param EventDispatcher $eventDispatcher
      * @return $this
      */
-    public function setEventDispatcher(Dispatcher $eventDispatcher)
+    public function setEventDispatcher(EventDispatcher $eventDispatcher)
     {
         $this->eventDispatcher = $eventDispatcher;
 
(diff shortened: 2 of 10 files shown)


src/Kris/LaravelFormBuilder/Form.php
```diff
@@ -1039,6 +1045,8 @@ public function validate($validationRules = [], $messages = [])
         $this->validator = $this->validatorFactory->make($this->getRequest()->all(), $rules, $messages);
         $this->validator->setAttributeNames($fieldRules['attributes']);
 
+        \Event::fire(new BeforeFormValidation($this, $this->validator));
```
kristijanhusak: Do not use Laravel facades in package. Inject all dependencies through constructor.

rudiedirkx:
Can do. Soon.

(Isn't a facade exactly the same as constructor injection? Both resolve from the container, right? Is there a specificity difference?)

kristijanhusak:
Great.

No, since this is a package, and should be set up like that so it can be used outside of Laravel.
Also, it's not testable this way.

rudiedirkx:
Now that I'm in the neighborhood, this seems off:

```
$this->app->singleton('laravel-form-builder', function ($app) {
    return new FormBuilder($app, $app['laravel-form-helper']);
});
$this->app->alias('laravel-form-builder', 'Kris\LaravelFormBuilder\FormBuilder');
```

If you bind a non-class, and then constructor inject it, a new one is created. I thought reversing the alias (and have no singleton) would fix that:

```
$this->app->alias('Kris\LaravelFormBuilder\FormBuilder', 'laravel-form-builder');
```

but it's still newly made every injection. How do you constructor inject a singleton? It seems strange to manually include `FormBuilder`'s constructor arguments when they could/should be auto injected.

This is how I use the form builder:

```
public function getEditOrganization(Request $request, FormBuilder $forms) {
```

so Laravel creates a new `FormBuilder`. If I ask the container for it later on (e.g. `app('laravel-form-builder')`) it creates a new one.

rudiedirkx:
I've added a property `eventDispatcher` to `FormBuilder` and `Form`.

I think the `laravel-form-builder` singleton looks strange, but maybe that's how it's supposed to work. I thought it's better to let the container resolve a service class, but it's still not a singleton then, and FormBuilder won't have the container itself.

rudiedirkx:
Added the event dispatcher everywhere now, including tests. This feels strange. Resolving from the container should be magic, right? =) Just tell it which class you want, and it'll do all the resolving. Now we have to manually 'fix' missing dependencies.

But tests and plain forms work again!

rudiedirkx:
Or maybe manual injection is exactly what you want, because changing dependencies will break 3rd party constructions that don't use the container's magic. Maybe? I'm pretty new to Laravel, so I'm guessing =)

rudiedirkx:
Added 2 tests to test events. Not sure how events are usually tested. This seems a bit hacky =)

src/Kris/LaravelFormBuilder/Events/BeforeFormValidation.php
```diff
@@ -0,0 +1,40 @@
+<?php namespace Kris\LaravelFormBuilder\Events;
```
kristijanhusak: There could maybe also be a AfterFormValidation event.

src/Kris/LaravelFormBuilder/Form.php
```diff
@@ -618,6 +630,19 @@ public function __get($name)
     }
 
     /**
+     * Set the Event Dispatcher to fire Laravel events
+     *
+     * @param EventDispatcher $eventDispatcher
+     * @return $this
+     */
+    public function setEventDispatcher(EventDispatcher $eventDispatcher)
```
kristijanhusak: We can remove this setter, and pull eventDispatcher form formHelper (add getter there)

src/Kris/LaravelFormBuilder/FormBuilderServiceProvider.php
```diff
@@ -30,7 +30,7 @@ public function register()
 
         $this->app->singleton('laravel-form-builder', function ($app) {
 
-            return new FormBuilder($app, $app['laravel-form-helper']);
+            return new FormBuilder($app, $app['laravel-form-helper'], $app['Illuminate\Contracts\Events\Dispatcher']);
```
kristijanhusak: I think `'Illuminate\Contracts\Events\Dispatcher'` can be simply replaced with `'events'`.

https://github.com/laravel/framework/blob/5.2/src/Illuminate/Foundation/Application.php#L1074

src/Kris/LaravelFormBuilder/Events/BeforeFormValidation.php
rudiedirkx: I don't know when validation happens. When the caller calls `isValid()` ? What if it's called multiple times? BeforeValidation is useful because that's the time to add logic to pass or fail the validation with custom logic. AfterValidation can't really do anything useful...

src/Kris/LaravelFormBuilder/Form.php
rudiedirkx: `FormHelper` doesn't have the dispatcher. `FormBuilder` does. Move to Helper and always ask Helper to fire events?, or take the dispatcher from Helper and ask it directly?

rudiedirkx:
* Will fix 'events' instead of FQN
* Need your input on `AfterFormValidation`
* Need your input on `setEventDispatcher` / dispatcher location

src/Kris/LaravelFormBuilder/Form.php
kristijanhusak: Ah, my bad. Thought you are passing it to FormHelper through constructor. You can leave it like this.

src/Kris/LaravelFormBuilder/Events/BeforeFormValidation.php
kristijanhusak: I think it shouldn't be called multiple times, there is no really a reason for that.
Some ppl may use this event to redirect back, or to log something, who knows.

It can be added around this line https://github.com/kristijanhusak/laravel-form-builder/blob/master/src/Kris/LaravelFormBuilder/Form.php#L1069.
Maybe something like this:

```php
$isValid = !$this->validator->fails();

$this->fire(new AfterFormValidation($isValid, $form, $this->validator));

return $isValid;
```

rudiedirkx: Done!

rudiedirkx:
Fixed `'events'` and `AfterFormValidation`. Didn't fix `setEventDispatcher`. Also merged new master, so it should be mergable again.

kristijanhusak:
This looks good now, thanks!

Issue #244, opened 2016-07-13
How to render form layout, in both form and form_child scenario


rudiedirkx:
I have an `OrganizationForm` that is used in 2 forms:

* creating an organization has an 'organization' fieldset with loads the `OrganizationForm` (with `ChildFormType`)
* editing an organization IS the `OrganizationForm`

It has a pretty sophisticated layout of fields, with internal fieldsets (only markup). I want to reuse that layout for both implementations: as a full form and as a child form. Is there a way?

I think what I need is the fields part of templates `form.php` and `child_form.php` to be the same, and configurable the same.

I can completely override `form.php` and `child_form.php` and maybe/almost share a `fields.php`, but the fields itself should be enough. Is there a good DRY way to only override the fields layout, for both the `form` and the `child_form` scenario?

kristijanhusak:
If you can post some code, it's easier for me to understand it that way.

rudiedirkx:
Alright. I'll try to be succinct.

I have 2 page forms:

* `OrganizationForm` contains the organization fields (there's a matching model `Organization`)
* `CreateOrganizationForm` contains 2 child forms: `OrganizationForm` and `UserForm`, because when you create a new organization, you also create an organization admin (user)

In both forms, the organization has the same many fields, in several fieldsets (just markup/layout). I'd like to share that field layout logic between both forms. I was hoping I could tell `OrganizationForm` to always render itself like `organization.partials.form`, but that apparently doesn't work for child forms, because they render as a field, not as a form.

This is the code I have now, to share the field layout logic between master form and child form:

`CreateOrganizationForm` includes both forms, and how to render the Organization part:

```
class CreateOrganizationForm extends Form {
  public function buildForm() {
    $this->add('organization', 'form', [
      'class' => OrganizationForm::class,
      'template' => 'organization.partials.child_form', // This is the child form, without <form> tags
    ]);

    $this->add('admin', 'form', [
      'class' => UserForm::class,
    ]);
  }
}
```

`organization/partials/child_form.twig` extends the normal `child_form` tpl (in Twig, sorry):

```
{% if showLabel and showField and options.wrapperAttrs %}
  <fieldset {{ options.wrapperAttrs|raw }}>
{% endif %}

{% if showLabel and options.label %}
  <legend {{ options.labelAttrs|raw }}>{{ options.label }}</legend>
{% endif %}

{% if showField %}
  {% include 'organization.partials.fields' with {'fields': options.children} %}
[...]

PR #256, opened 2016-08-02
#244 +FieldsContainerContract, +form_rows(), +form option

rudiedirkx:
PR for option 1 of #244 

* Added interface `FieldsContainerContract` to typehint a form-like.
* Added 'form' (the ChildFormType field object) to the options. I'd like it to be in the root vars, not in the options, but FormField has a very strict list of view params.
* Added `form_rows()` to render several form fields at once, by field names instead of multiple `form->getField()->render()`s

I like option 2 better (a `fields.php` tpl), but this is still a good addition IMO.

### DIFF
--- a/src/Kris/LaravelFormBuilder/Fields/ChildFormType.php
+++ b/src/Kris/LaravelFormBuilder/Fields/ChildFormType.php
@@ -1,8 +1,9 @@
 <?php  namespace Kris\LaravelFormBuilder\Fields;
 
+use Kris\LaravelFormBuilder\FieldsContainerContract;
 use Kris\LaravelFormBuilder\Form;
 
-class ChildFormType extends ParentType
+class ChildFormType extends ParentType implements FieldsContainerContract
 {
 
     /**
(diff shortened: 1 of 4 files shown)


kristijanhusak:
Why is interface called `FieldsContainerContract` ? Maybe something like FormContract or FormInterface would be a bit more suitable.

rudiedirkx:
Because it contains fields, but doesn't have any other Form properties. I'm not a huge fan either, but `FormContract` might be confusing, because it's not a real form. Or do we just pretend it is, since it has a `__call` forward..?

rudiedirkx:
Maybe this should go another direction. The only thing I missed (aside from `fields.php` tpl, but later) is a `$form` variable in the child form tpl. I don't like it in options anyway. Can we just add a form variable to this field's render variables somehow? The interface might be overkill...

kristijanhusak:
Yeah i think it's a bit overkill. Just pass the form with options, that should do it.

rudiedirkx:
If we'd add `form` and `field` in `FormField::render()`, you have all you could want in the tpl, BUT that might be too much. That's for you to decide. A themer could break stuff with a full `$form` or `$field` object. I usually don't mind: the more details the better, but some people do mind.

I'll change the PR to include `$form` (`$this->parent`) and `$field` (`$this`) in `FormField::render()`. That will solve the immediate styling issues. Part 2 of #244 (a `fields.php` tpl) will be the ultimate boss stuff, but later.

If I hurry this PR, will it be included in the next release? I'd reeaally like that =)

rudiedirkx:
(Also, any more thoughts on #244 are much appreciated, so I know what to focus on.)

rudiedirkx:
Closing in favor of https://github.com/kristijanhusak/laravel-form-builder/pull/262

PR #255, opened 2016-08-01
Adds attr.append_class option to fields

theshaunwalker:
Adds the ability to use `attr.append_class` to append to the `attr.class` value instead of having to overwrite it. This allows you to use global defaults and not have to hard code them if you just want to add more classes.

PR #257, opened 2016-08-06
Add ability to append to class attribute for wrapper, label and field

koenvu:
Originally proposed by @theshaunwalker in #255 

> Adds the ability to use attr.append_class to append to the attr.class value instead of having to overwrite it. This allows you to use global defaults and not have to hard code them if you just want to add more classes.

I've rewritten it a bit, made sure that `class_append` would not show up as an actual attribute and wrote tests covering appending to wrapper, label and field.


kristijanhusak:
Thanks, its great!

PR #258, opened 2016-08-06
Redirect when verification fails

koenvu:
Since the following pattern happens for almost every form I'd say, we can create something that is a little less verbose. This code in the controller:

```
$form = magicallyCreateSomeForm();

if (!$form->isValid()) {
    return redirect()->back()->withErrors($form->getErrors())->withInput();
}
```

Would become:

```
$form = magicallyCreateSomeForm();

$form->redirectIfNotValid();
```

Or even this if you don't need really the need the form object after validating:

```
magicallyCreateSomeForm()->redirectIfNotValid();
```

I figured the `Form` class would be a good spot for this, although one could argue that a `redirectIfNotValid($form)` function on the `FormBuilderTrait` might work as well.

A test has been written to ensure a redirect with input and errors is returned when verification fails. I've been using this in one of my projects, and I really like that it cleans up my controller just a little bit more.

### DIFF
--- a/src/Kris/LaravelFormBuilder/Form.php
+++ b/src/Kris/LaravelFormBuilder/Form.php
@@ -1056,10 +1056,18 @@ class Form
         return array_merge($fieldRules['rules'], $overrideRules);
     }
 
-    public function redirectIfNotValid()
+    public function redirectIfNotValid($destination = null)
     {
         if (! $this->isValid()) {
-            throw new HttpResponseException(redirect()->back()->withErrors($this->getErrors())->withInput());
+            $response = redirect($destination);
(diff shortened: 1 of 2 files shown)


kristijanhusak:
Thanks. This looks great. I would maybe allow passing in an url, in case someone wants to redirect somewhere else, and not back. If it's not provided, do the `back()` action.

koenvu:
Wasn't able to properly fully test the `back()` without mocking or muddying up the test as far as I can see, but I think this should be good.

kristijanhusak:
Tests are fine, thanks for PR!

PR #261, opened 2016-08-11
Fix adding required class to labels when client validation is disabled

koenvu:
The `required` class was not added when client side validation is not enabled. I think it makes sense to add the `required` class to labels even when validation is not enabled. My mistake in #209 😉 

I also changed the test names I wrote back then to be a little more descriptive and added a check for the `required` attribute as well.

If possible, can we create a new version from this soon? I'd love to push this fix to production without having to use my fork as the source for the package. Thanks so much!


kristijanhusak:
Yeah this makes sense.

I'll publish new version soon, when i merge other PRs.

kristijanhusak:
@koenvu new version is published with this change, you can update.

PR #262, opened 2016-08-11
#244 +form & field vars in field render, +form_rows() helper

rudiedirkx:
Follow-up of https://github.com/kristijanhusak/laravel-form-builder/pull/256 which was too much.

With these 2 additions, you can make a reusable fields tpl that renders several fields at once:

In `form.php`:

    {{ form_rows(form, ['name', 'age']) }}

In `child_form.php`:

    {{ form_rows(field, ['name', 'age']) }}

because `form` is the form and `field` is the child form, kinda.

I don't like the `$form` argument in `form_rows()` though. It can be a `Form` or a `ChildFormType`, and both work because of `ChildFormType::__call()`, but you can't type hint that...

### DIFF
--- a/src/Kris/LaravelFormBuilder/Fields/ChildFormType.php
+++ b/src/Kris/LaravelFormBuilder/Fields/ChildFormType.php
@@ -151,6 +151,15 @@ class ChildFormType extends ParentType
     }
 
     /**
+     * @inheritdoc
+     */
+    protected function getRenderData() {
+        $data = parent::getRenderData();
+        $data['child_form'] = $this->form;
+        return $data;
+    }
+
+    /**
      * @param $method
      * @param $arguments
      *
--- a/src/Kris/LaravelFormBuilder/Fields/FormField.php
+++ b/src/Kris/LaravelFormBuilder/Fields/FormField.php
@@ -161,11 +161,11 @@ abstract class FormField
             $showError = $this->parent->haveErrorsEnabled();
         }
 
+        $data = $this->getRenderData();
+
         return $this->formHelper->getView()->make(
             $this->getViewTemplate(),
-            [
-                'form' => $this->parent,
-                'field' => $this,
+            $data + [
                 'name' => $this->name,
                 'nameKey' => $this->getNameKey(),
                 'type' => $this->type,
@@ -175,6 +175,15 @@ abstract class FormField
                 'showError' => $showError
             ]
         )->render();
+    }
+
+    /**
+     * Return the extra render data for this form field, passed into the field's template directly.
+     *
+     * @return array
+     */
+    protected function getRenderData() {
+        return [];
     }
 
     /**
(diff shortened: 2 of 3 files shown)


kristijanhusak:
Yeah that's kind of ugly. Let's leave it like this for now.

Also, i'm not so sure about sending whole Field (`$this`) to the view. I thought only sending childForm for ChildFormType. Do you need Field instance in view for some particular reason?

rudiedirkx:
> Do you need Field instance in view for some particular reason?

Only for the child form type, for sending to a homemade fields tpl.

> Let's leave it like this for now.

Leave it how? Like this PR, or don't do the PR?

Maybe even better: make a FormField method to pass field-type-specific vars into the field renderer. So `FormField::render()` takes care of the render logic, but uses `SomeType::renderVars()` for specifics, and currently only ChildFormType would implement that to add `'form' => $this`.

I like it. You?

rudiedirkx:
I do like the field instance better as `$field` or `$form` though, not as semi-hidden option in `$options['field']`.

kristijanhusak:
> Only for the child form type, for sending to a homemade fields tpl.

What exactly u need from it? I would like to send only the child form instance, not the whole field ($this).

By leave, i mean to use this PR as it is.

We could add that renderVars method, that will only return some things to pass. Could be empty array by default. Maybe something like this:

```php
        return $this->formHelper->getView()->make(
            $this->getViewTemplate(),
            array_merge([
                'name' => $this->name,
                'nameKey' => $this->getNameKey(),
                'type' => $this->type,
                'options' => $this->options,
                'showLabel' => $showLabel,
                'showField' => $showField,
                'showError' => $showError
            ], $this->renderVars())
        )->render();
```

rudiedirkx:
That's what I thought. And I missed something stupid, that I think you meant (pass in the actual child form (the `Form`), not the child form type field (the `ChildFormType`)) so I did that too. Now `form_rows()` always receives a `Form`, either the parent form of the child form type's real form. Eh, yeah, just look at the diff now, it's much easier.

rudiedirkx:
My `resources/views/organization/child_form.php` looks like this now:

```
<?php if ($showField): ?>
	<?php $form = $child_form; include 'fields.php'; ?>
<?php endif; ?>
```

And `fields.php`:

```
<fieldset>
	<legend>Billing details</legend>
	<?= form_rows($form, ['billing_contact_person', 'billing_address', 'billing_email', 'billing_phone']) ?>
</fieldset>
```

and in case of the form as the main/parent form, `form.php` is rendered:

```
<?php if ($showFields): ?>
	<?php include 'fields.php'; ?>
<?php endif; ?>
```

which already has the correct `$form` variable.

And that works perfectly.

I still like a native `fields.php`, but this is a great interim solution IMO.

kristijanhusak:
Ok, great. Thanks!

rudiedirkx:
Schweet! Now I have to plan some form refactor time soon. The boss will like that.

PR #263, opened 2016-08-11
#246 gh-pages update for Form::getFieldValues()

rudiedirkx:
* Use `$form->getFieldValues()` instead of `$request->all()` because it's safer
* Point to `$form->redirectIfNotValid()` because it's cool


kristijanhusak:
Thanks man!

PR #265, opened 2016-08-12
No getFieldValues() for (submit) buttons

rudiedirkx:
I forgot one. Submit buttons don't need values. If you `getFieldValues()` with the optional argument to its default (TRUE), it adds NIL values for 'missing' input like a submit button.


kristijanhusak:
Ah, yeah. i'll publish minor fix later.

PR #272, opened 2016-08-18
Alter field values

rudiedirkx:
Added

* `alterFieldValues()` to `Form` to alter that form's values
* `alterFieldValues()` to `FormHelper` to call that alter and recurse through fields
* `alterFieldValues()` to `ChildFormType` to act on and recurse through child forms

Three different functions with the same name... Might not be the best.

With test!

There's still something missing and it might be part of this: form specific validation. If a child form has an image and another form has validation errors, the child form wants to add a validation message to reupload the image. The parent form could do this in `isValid()`, but that's not called for child forms... Ideas?

### DIFF
--- a/src/Kris/LaravelFormBuilder/Fields/ChildFormType.php
+++ b/src/Kris/LaravelFormBuilder/Fields/ChildFormType.php
@@ -58,6 +58,14 @@ class ChildFormType extends ParentType
     }
 
     /**
+     * Allow form-specific valid alters
+     */
+    public function alterValid(Form $mainForm, &$isValid)
+    {
+        $this->parent->getFormHelper()->alterValid($this->form, $mainForm, $isValid);
+    }
+
+    /**
      * @return mixed|void
      */
     protected function createChildren()
--- a/src/Kris/LaravelFormBuilder/Fields/ChildFormType.php
+++ b/src/Kris/LaravelFormBuilder/Fields/ChildFormType.php
@@ -50,6 +50,14 @@ class ChildFormType extends ParentType
     }
 
     /**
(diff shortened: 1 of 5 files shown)


rudiedirkx:
Ah crap. `alterValid()` doesn't have access to the correct form's field values... Every `getFieldValues()` returns the top form's values, sort of, so that's not useful for a nested form that doesn't know its nesting location.

The nicest would be to 'fix' `getFieldValues()` to make it callable from any nested form. I'm gonna think about that...

rudiedirkx:
This should do it.

* `getFieldValues()` now works for child forms
* Form has an optional `alterValid()` to alter the validation result and add messages

kristijanhusak:
Why do you need this alter value method? Give me an example pls.

rudiedirkx:
`alterFieldValues()`:

* checkboxes + none selected = missing value, but should be an empty array
* dropdown of user reference where none selected = empty string or 0, but should be NULL

`alterValid()`:

* Uploading a file + other validation error = must reupload file, because file input can't have default value. A message to tell the user that would be nice.
* Validation on a difficult combination of 3 or 4 form fields to make $isValid = FALSE.

Stuff like that is part of the form (validation/values), not the controller, so it'd be nice if the form could do that. Overriding `getFieldValues()`/`isValid()` isn't enough because that's not called for child forms and child child forms etc.

I don't think there's another way to do both things (alter valid and alter values) currently.

kristijanhusak:
Great. Everything looks fine. Thanks!

rudiedirkx:
Awesome! It's almost a perfect the best form builder in the world!

PR #274, opened 2016-08-20
Bugfix in getFieldValues() for CollectionType and RepeatedType

rudiedirkx:
More `getFieldValues()` bugs.

* `RepeatedType` skipped the repeated value, because it wasn't in the attributes list
* `CollectionType` added the **entire** data blob for that element, no matter the format, so now it collects children like ChildFormType does

I checked all ParentType components and I think this is it. The rest are radio buttons etc, so they should be included as 1, not per child element.

Aside: I expected CollectionType to validate with `*`: `locations.*.name`, `locations.*.street` etc, but it validates only explicitly present elements, so `locations.0.street`, `locations.1.street` etc. How would you validate a new location added with JS? It's not in the validation list. (And now it's not in the attributes list, so it won't be included in `getFieldValues()`).

### DIFF
--- a/src/Kris/LaravelFormBuilder/Fields/CollectionType.php
+++ b/src/Kris/LaravelFormBuilder/Fields/CollectionType.php
@@ -61,6 +61,15 @@ class CollectionType extends ParentType
     /**
      * @inheritdoc
      */
+    public function getAllAttributes()
+    {
+        // Collect all children's attributes.
+        return $this->parent->getFormHelper()->mergeAttributes($this->children);
+    }
+
+    /**
+     * @inheritdoc
+     */
     protected function createChildren()
     {
         $this->children = [];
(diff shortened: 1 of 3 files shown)


rudiedirkx:
CollectionType works even cooler than I thought! No `*` validation, but dynamic elements. Very cool. Works perfectly with validation errors. My bad =)

rudiedirkx:
But how do you make a simple collection (textfields) required without making all textfields required? Adding `required` and/or `min:1` to the collection element validates the array with empty element. Adding `required` to the textfields (in `options`) makes all textfields required. Must be custom validation? Same for child form collections actually: elements inside the child form are required IF that child form is 'used', but you can leave it completely empty, which shouldn't validate any inside rules. Would work for simple collections too...

kristijanhusak:
I think that's the most used use case, where you need it to be validated if it's used. For collection empty array validation, i'm not really sure. I guess custom validation must be added, haven't done it tho.

Thanks for the fix.

PR #291, opened 2016-10-19
Dont overwrite $options['real_name']

theshaunwalker:
If you specified 'real_name' in the options when adding a field it didnt matter because it was overriden here.

Just adds a simple check so it doesnt overwrite the 'real_name' value if the user has already set it.

### DIFF
--- a/src/Kris/LaravelFormBuilder/Form.php
+++ b/src/Kris/LaravelFormBuilder/Form.php
@@ -941,7 +941,10 @@ class Form
      */
     protected function setupFieldOptions($name, &$options)
     {
-        $options['real_name'] = $name;
+        // Only set real_name if not already specified
+        if (!isset($options['real_name'])) {
+            $options['real_name'] = $name;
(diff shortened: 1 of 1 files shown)


kristijanhusak:
Why would you set custom `real_name` ? I'm using that internally.

theshaunwalker:
I was adding multiple checkbox fields for days on a form `days[mon]` `days[tue]` etc.

This made their array keys `days[mon]` and using `->getField('days[mon]')` fails because PHP doesnt seem to process square brackets in an array key. So I needed to convert the key to snake case so I could grab it properly.

kristijanhusak:
The `real_name` in that case should be `mon`, `tue`, etc.
How did you create those multiple checkboxes? Separately or with Choice type?

theshaunwalker:
Ah, separately, I completely forgot about the multiple option on choice. My bad, brain fart haha.

Disregard this then.

PR #293, opened 2016-10-26
support horizontal form

tantam:
add some methods to check form if it is horizontal form and add some
bootstrap class to html form and fields

P/s: This is the first time I pull a request

### DIFF
--- a/src/Kris/LaravelFormBuilder/Form.php
+++ b/src/Kris/LaravelFormBuilder/Form.php
@@ -45,8 +45,10 @@ class Form
      */
     protected $formOptions = [
         'method' => 'GET',
-        'url' => null
+        'url' => null,
     ];
+
+    protected $formStyle = 'default';
 
     /**
      * Additional data which can be used to build fields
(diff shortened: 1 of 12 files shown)


tantam:
example code: https://gist.github.com/tantam/e1190d5f7a65d78094ba382193dedb1d
screen shot 2016-10-26 at 5 16 51 pm

votanlean:
great!

kristijanhusak:
Thanks for the PR. General idea is ok, and code works, but there are bunch of issues with code structure. I'll summarize them here so i don't leave too many comments on the code:

1.  Fix all code to be according to PSR-2 (spaces around curly braces, indentation, line limitation, etc.)


2. Fix long if statements (example [this one](https://github.com/kristijanhusak/laravel-form-builder/pull/293/files#diff-7f17d71f56dd1175c49875c83e9adc04R362)) - It's completely unreadable. Try to move some of those to a variable.

3. Adapt views structure. Do not write `echo "</div>"`, pull that out to plain html, like other ifs.

4. Do not use `app()` helper at all, especially not in the views. Prepare wrapper attributes in the Field class, and pass them to view, for example `rightWrapperAttrs`.

I think that's it for now.

tantam:
Thanks for you feedbacks, I will fix them

jojonas:
@tantam What's the progress in this MR? I would also be interested in this functionality.

tantam:
sorry, I don't have time for it so I close this PR

PR #314, opened 2017-02-06
Improved DocComments, params especially their type, returns, line endings

StefanPrintezis:
Fixed some inconsistencies 
* Always using implode instead of 'join' once
* return types
* . on the end of a comment
* { class opening on new row. (same for functions)
* inheritdoc where possible
* properties don't require naming the $foo in the comment.
* Comments are no longer questions. 
* @params have types now.
* Naming boolean -> bool


kristijanhusak:
Thanks!

PR #317, opened 2017-02-10
Test lowest/stable

barryvdh:
Test lowest versions of the required packages also


kristijanhusak:
Closing since it's outdated. If you think this should be still done, let me know.

PR #310, opened 2017-01-28
Added Laravel 5.4 compatibility

mauricius:
Fixes #307

PR #318, opened 2017-02-10
throwResponse in Laravel 5.3+, old response otherwise

barryvdh:
Test lowest versions of the required packages also. The Exception changed for 5.4 support (#310), but broke old support.

This check for pre-5.3 version and throws the old exception. New one uses the throwResponse trait added in 5.2.32; https://github.com/laravel/framework/blob/master/CHANGELOG-5.2.md#v5232-2016-05-17

### DIFF
--- a/src/Kris/LaravelFormBuilder/Form.php
+++ b/src/Kris/LaravelFormBuilder/Form.php
@@ -2,10 +2,10 @@
 
 namespace Kris\LaravelFormBuilder;
 
+use Illuminate\Foundation\Application;
 use Illuminate\Contracts\Events\Dispatcher as EventDispatcher;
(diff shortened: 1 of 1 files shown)


SebTM:
This would be important as we currently have this issue too!

pimlie:
The version comparison is not correct, there is no need to substr the current version when you use `version_compare`. Also with the current implementation you get problems once laravel reaches version 5.30 (maybe unlikely, but still).

You could just use:
```php
if(version_compare(Application::VERSION, '5.3.0', '<'))
```

kristijanhusak:
Is this still an issue? I'm not into php development lately.

Issue #279, opened 2016-09-07
Access model in child form in collection


rudiedirkx:
In a child form in a form, you access the child form's data model. It's not the actual model object in `$this->model`, but the parent's data, and by following the `$this->name` you can get the actual model. That's weird, but probably can't be changed. In a child form in a collection, there's no model at all, but there is `$this->data`, with the almost-same weird format, but the entire data, not just the relevant item.

What's a dependably way to get the current (child) form's data? I know it's always a model, but I don't know where it is. The default values work perfectly, somehow.
- `$this->model` is sometimes a model, and sometimes an array with the right key
- `$this->data` is sometimes an array, and sometimes a model collection

This might be a stupid question, but why isn't `$this->model` always the relevant model? If the default values are filled correctly, it must know the right model...

I tried debugging, but I got lost =)

DanielaSp:
Hi, I'm having the same issue:
```php
//List Form
public function buildForm()
{
  $list = $this->model; //this is correct
  $this
    ->add('name', 'text', [
        'rules' => 'required|min:5'
    ])
    ->add('items', 'collection', [
     'type' => 'form',
     'options' => [    // these are options for a single type
         'class' => 'App\Forms\ItemForm',
         'label' => false,
         'fields' => $this->getData('fields'),
     ],
   ])
   ->add('submit', 'submit', ['label' => 'Create list']);
} 

//ItemForm
public function buildForm()
{
  $item = $this->getModel(); //this is still List model, while I'm looking for the Item model
  $this 
    ->add('id', 'text'); //this is correct of course
}
```
How do I get the Item model in a collection of child forms?

rudiedirkx:
Fixed in 3566122 ?

PR #325, opened 2017-02-16
Fix child models not being bound correctly in collection forms

njbarrett:
This is my attempt to fix this issue https://github.com/kristijanhusak/laravel-form-builder/issues/279

When you make a collection of child forms, the bound model is the parent's model rather than the child's model.

For example, if I have a UserProfileForm with a collection of child forms called ProductForm, when accessing the getModel() of the product form I get back the user model from UserProfileForm when it should be returning the product model.

I also attempted to write a test for this.


kristijanhusak:
This looks ok. I haven't tested it manually, but since all previous tests passed, i believe it should be ok. Thanks!

Issue #319, opened 2017-02-10
Change which form class to use in `plain()`


rudiedirkx:
`Kris\LaravelFormBuilder\FormBuilder::plain()` creates a simple `Kris\LaravelFormBuilder\Form`, which is hardcoded. I have created a custom base form that I use everywhere, and I'd like `plain()` to use that form. I can replace `FormBuilderServiceProvider` and `FormBuilder` and `plain()`, but that seems like overkill. Maybe my own service provider could tell the `laravel-form-builder` service to load another plain class?

rudiedirkx:
This is harder than expected, because there are so many ways to create a FormBuilder instance. First of all, I think this is the wrong way round:

```
$this->app->singleton('laravel-form-builder', function ($app) {
	return new FormBuilder($app, $app['laravel-form-helper'], $app['events']);
});
$this->app->alias('laravel-form-builder', 'Kris\LaravelFormBuilder\FormBuilder');
```

I think the service should be registered by its class name, and then aliased to a pretty name.  If you do that, you can inject it:

```
public function getTranslate(Request $request, FormBuilder $forms) {
```

but then there's the extension. I want to add a few form builder methods, which creates a new class name, which is not the one that was registered. If one place type hints mine, and another place type hints yours, there are 2 form builders floating around. Aliasing doesn't help either way. The only solution seems to be registering my own form builder with my own name, and use that everywhere, and hope nobody else ever type hints yours.

Any brilliant ideas? Maybe it's becoming a Laravel service container issue and I just don't know how that works...

rudiedirkx:
Or maybe create an alias from my FormBuilder to your FormBuilder (or the other way around)? So if someone type hints your FormBuilder, they get mine, which is still valid, because it extends yours.

That doesn't seem to work, but it totally should! That's Laravel's fault. It doesn't check aliases for typehint injection. (I'm guessing.)

rudiedirkx:
laravel.io forum post: https://laravel.io/forum/02-19-2017-alias-service-class-to-service-class

rudiedirkx:
Okay, ignore my service provider ramblings. There is a PR coming for the actual issue.

rudiedirkx:
Fixed in cf56836

PR #328, opened 2017-02-19
#319 Configurable plain form class

rudiedirkx:
Allows any service provider to change the plain form class to any extension of the default Form class.


kristijanhusak:
Thanks!

PR #334, opened 2017-02-28
Use primaryKey of model as default property_key

pimlie:
Currently there is a fixed default value of 'id' for the property_key name. This is because the default primaryKey column in Laravel is also 'id'. But as we already instantiate the entity, this commit just requests from the entity what its key name is.


kristijanhusak:
Nice, thanks!

PR #336, opened 2017-03-14
ChoiceType: Deal with child options

schursin:
Now we can set up child options, for example **size** or **VueJS attributes**.

PHP:

```
$this->add('languages', 'choice', [
    'choices' => ['en' => 'English', 'fr' => 'French'],
    'multiple' => true
]);
```

Blade:

`{!! form_row($form->languages, ['attr' => ['size' => 10, '@input' => 'update']]) !!}`

Output before:

`<select class="form-control" multiple="1" id="languages[]" name="languages[]">`

Output after:

`<select class="form-control" multiple="1" size="10" @input="update" name="languages[]">`

### DIFF
--- a/src/views/choice.php
+++ b/src/views/choice.php
@@ -10,7 +10,7 @@
 
 <?php if ($showField): ?>
     <?php foreach ((array)$options['children'] as $child): ?>
-        <?= $child->render(['selected' => $options['selected']], true, true, false) ?>
+        <?= $child->render(['attr' => $options['attr'], 'selected' => $options['selected']], true, true, false) ?>
     <?php endforeach; ?>
 
     <?php include 'help_block.php' ?>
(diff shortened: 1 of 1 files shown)


kristijanhusak:
You should be able to achieve that by passing options through `choice_options`.
http://kristijanhusak.github.io/laravel-form-builder/field/choice.html

schursin:
Now:

`{!! form_row($form->languages, ['child_options' => ['attr' => ['size' => 10, '@input' => 'update']]]) !!}`

kristijanhusak:
@schursin no, my point is that you should be able to already do that, since merging is done in the class here https://github.com/kristijanhusak/laravel-form-builder/blob/master/src/Kris/LaravelFormBuilder/Fields/ChoiceType.php#L101

schursin:
@kristijanhusak checked that, not working

Template:

`form_row($form->languages, ['child_options' => ['attr' => ['size' => 10, '@input' => 'update']]])`

HTML:

`<select class="form-control" multiple="1" id="languages[]" name="languages[]">`

**size** attribute is missing :(

kristijanhusak:
@schursin thanks!

Issue #340, opened 2017-03-29
Extensible RulesParser


rudiedirkx:
For some rules, the form builder automatically adds HTML validation attributes. Perfect. But not for the `regex` rule. It adds `pattern` attributes sometimes, but not for the `regex` rule. And there might be more (custom) validations that could auto-add attributes.

If you'd make the `RulesParser` extensible, anyone could add this 'rule to attribute' logic. Maybe a `createRulesParser()` method in the `FormHelper`, because the form helper is already extensible.

barryvdh:
The reason pattern is not always added, is because it's not always the same in HTML5 as php, so can be a bit confusing.

rudiedirkx:
Fair enough. What about making it extensible so I can choose to use JS-safe regexes only?

barryvdh:
Makes sense to me, but not my package :)

rudiedirkx:
Yeah Barry! What are you doing here anyway :stuck_out_tongue:

Kris is better with the PRs, so I'll do that soon.

I totally missed your link to #144 btw. Good context. Thanks for clarification.

PR #342, opened 2017-03-30
+Extensible RulesParser

rudiedirkx:
From #340 

This allows the developer to extend the RulesParser to add their own 'rule to attribute' logic.


kristijanhusak:
Maybe i'm a bit confused, but how would you extend it with this code? Can you give an example?

kristijanhusak:
Nevermind, figured it out.

kristijanhusak:
Thanks for the PR.

rudiedirkx:
It's a little convoluted, but extending the FormHelper is pretty easy. Maybe a factory closure (or what does Laravel call them) would have been better, but meh.

PR #343, opened 2017-04-05
Fix HttpResponseException namespace

MikeMnD:


MikeMnD:
I'll close this since we found it's Laravel 5.4 where the namespaces is changed. And our project is on 5.3

PR #344, opened 2017-04-05
Bug Fix - ButtonGroupType.php

jayjfletcher:
Button group field type was not working out of the box due to a non-existent path being set. Fixed the button group template path to match the rest of the types.


kristijanhusak:
Thanks!

PR #345, opened 2017-04-06
Add ability to automatically validate Form classes

mpociot:
This PR adds the ability to automatically validate form objects when they get resolved through Laravels IOC container.

The behavior is similar to the FormRequest validation, so you only need to typehint the form object in your controller. If the validation fails, the form will automatically trigger the `redirectIfNotValid` method.

Usage:

## Let your form class use the ValidatesWhenResolved trait
```php
<?php

use Kris\LaravelFormBuilder\Traits\ValidatesWhenResolved;

class TestForm extends Form
{
    use ValidatesWhenResolved;

    public function buildForm()
    {
        $this->add('name', 'text', ['rules' => ['required', 'min:3']]);
    }

}
```

## In your controllers POST method, typehint the TestForm

```php
<?php

class TestController
{

    public function validate(TestForm $form)
    {
        // Everything in here is automatically valid!
    }
    
}
```


kristijanhusak:
@mpociot Thanks for the PR! I'll do the refactoring that i mentioned in comments.

PR #346, opened 2017-04-07
Bug-fix: CollectionType.php

unfalln:
Use input() to correctly return the data from the current request so that, if it contains multiple rows, those rows can be read in as children of the Collection field.

### DIFF
--- a/src/Kris/LaravelFormBuilder/Fields/CollectionType.php
+++ b/src/Kris/LaravelFormBuilder/Fields/CollectionType.php
@@ -77,7 +77,7 @@ class CollectionType extends ParentType
         $this->children = [];
         $type = $this->getOption('type');
         $oldInput = $this->parent->getRequest()->old($this->getNameKey());
-        $currentInput = $this->parent->getRequest()->get($this->getNameKey());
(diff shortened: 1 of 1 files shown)


unfalln:
I will try to resubmit at some stage with a proper unit test to show what's going on.
In the meantime, I believe this is an issue where I am trying to submit a nested sub-form and the Request->get() function is not returning the same values as Request->input(), perhaps due to the array-style naming of get/post values. Depending on what I find out I may have to instead submit this to illuminate or somewhere else.

kristijanhusak:
@unfalln  what is really a difference between get and input methods?

unfalln:
I believe what was happening was, when I retrieved a POST with a SubFormType field that contains another form with a CollectionType field, the get() method could not correctly reference the incoming post data, where the input method could. This meant that, when I used added items to the CollectionType field on the client in javascript then ajaxed them back to the server, the CollectionType would only count the number of items in the defaultValue of the field and, later, only retrieve one row from post.

kristijanhusak:
@unfalln  ok, thanks. Merging this in.

PR #348, opened 2017-04-20
Check if custom field already defined on rebuildForm

alamcordeiro:
Prevent ```Custom field [name] already exists on this form object.``` error on addCustomField and rebuildForm.


kristijanhusak:
@alamcordeiro  Thanks for the PR. I'll refactor the part from comments. Merging this in.

PR #350, opened 2017-05-04
Adds form_fields helper

mjerwin:
Adds a new `form_fields()` helper function which renders the contents of the form without the opening and closing `<form>` tags.


kristijanhusak:
Thanks!

PR #351, opened 2017-05-15
Ignore exception thrown in FormField when adding classes

andersevenrud:
In my current project I made a Laravel Artisan Command that generates Migration files from a set of `Form` classes. This did not work as I was getting an `Exception` because no `Session` was available when the form builder gets to the point of assigning classnames to fields based on input from a `Request`.

This commit simply ignores this exception.

I don't know this is the correct way to do this however; so if you have any comments, please let me know and I'll figure out a better way (if this at all is accepted).


kristijanhusak:
@andersevenrud This is not really a good solution. You should somehow make session available in the request, this can swallow some errors that should show up.

PR #356, opened 2017-05-30
ParentType not pushing options to children

pimlie:
The ParentType FormField's only copies the options supplied on FormField creation to their children. If you call 'setOption/setOptions' after creating e.g. an 'EntityField' the option is never passed to the child. This commit changes that behaviour by calling 'setOption/setOptions' on all children as well. See added test in EntityType for an use-case.


kristijanhusak:
Thanks!

PR #364, opened 2017-07-13
Include key in checkbox name when using multiple choice type.

Luukvdo:
The following form:
```<?php

namespace App;

class TestForm extends Form
{
    public function buildForm()
    {
        $this
            ->add('choice', 'choice', [
                'multiple' => true,
                'expanded' => true,
                'choices' => [
                    'A',
                    'B',
                    'C',
                    'D',
                    'E',
                    'F'
                ]
            ])
            ->add('submit', 'submit');
    }
}
```
Results in:
Form

Selecting A & F results in the following array after submitting:
Submitted form
As you can see the indices being returned are 0 & 1, while 0 & 5 are selected.

I have changed the code so indices are included. I'm not sure if this is backwards compatible however, so please let me know.

### DIFF
--- a/src/Kris/LaravelFormBuilder/Fields/ChoiceType.php
+++ b/src/Kris/LaravelFormBuilder/Fields/ChoiceType.php
@@ -94,8 +94,6 @@ class ChoiceType extends ParentType
      */
     protected function buildCheckableChildren($fieldType)
     {
-        $multiple = $this->getOption('multiple') ? '[]' : '';
-
         foreach ((array)$this->options['choices'] as $key => $choice) {
             $id = str_replace('.', '_', $this->getNameKey()) . '_' . $key;
             $options = $this->formHelper->mergeOptions(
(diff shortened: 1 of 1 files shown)


Luukvdo:
@kristijanhusak Have you had any change to look at this? Thanks in advance

kristijanhusak:
@Luukvdo you get that from Laravel Request, or from the Form get field values?

Luukvdo:
@kristijanhusak I'm using the `getFieldValues` function on the Form.
```php
$form = $formBuilder->create(TestForm::class);

dd($form->getFieldValues());
```

kristijanhusak:
@Luukvdo what's the structure of that value if you take it from the request?

Luukvdo:
It turns out I changed the checkbox view at some point from
`<?= Form::checkbox($name, $options['value'], $options['checked'], $options['attr']) ?>`
to
`<?= Form::checkbox($name, null, $options['checked'], $options['attr']) ?>`
This resulted in the indices not being there and therefore in incrementing indices instead of the correct ones.
I'm not sure why I changed it, but after changing it back `$form->getFieldValues()` is returning the correct result.

I'll close this pull request now, thanks for looking at it anyway.

PR #369, opened 2017-08-11
fix entity relation

veneliniliev:
work with dimsav/laravel-translatable

$this->add('coutry_id', 'entity', [
            'class' => Country::class,
            'property' => 'name'
        ]);

when `name` property in countries_translation table.

->pluck(...) make query without relations :(


kristijanhusak:
You can use `query_builder` option to return a custom result. Something like this should work for you:

```php
$this->add('coutry_id', 'entity', [
    'class' => Country::class,
    'property' => 'name',
    'query_builder' => function ($country) {
        return $country->get();
    }
]);

```

PR #379, opened 2017-09-06
If a form field is not required, it won't have any client validation enabled

koichirose:
Just checked this with inputs and textareas.

This produces nothing:
```
'rules' => 'max:25',
```

This produces `required="required" maxlength="25"`
```
'rules' => 'required|max:25',
```
I very quickly created this PR, but I'm not sure it's actually a bug.

What do you think?

### DIFF
--- a/src/Kris/LaravelFormBuilder/Fields/FormField.php
+++ b/src/Kris/LaravelFormBuilder/Fields/FormField.php
@@ -295,11 +295,13 @@ abstract class FormField
 
             if ($this->parent->clientValidationEnabled()) {
                 $this->setOption('attr.required', 'required');
+            }
+        }
 
-                if ($parsedRules) {
-                    $attrs = $this->getOption('attr') + $parsedRules;
-                    $this->setOption('attr', $attrs);
-                }
+        if ($this->parent->clientValidationEnabled()) {
+            if ($parsedRules) {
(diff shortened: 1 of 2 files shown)


src/Kris/LaravelFormBuilder/Fields/FormField.php
```diff
@@ -295,11 +295,13 @@ protected function prepareOptions(array $options = [])
...
             if ($this->parent->clientValidationEnabled()) {
                 $this->setOption('attr.required', 'required');
+            }
+        }
 
-                if ($parsedRules) {
-                    $attrs = $this->getOption('attr') + $parsedRules;
-                    $this->setOption('attr', $attrs);
-                }
+        if ($this->parent->clientValidationEnabled()) {
```
kristijanhusak: You can move `$parsedRules` into this if to avoid double nesting:
```php
if ($this->parent->clientValidationEnabled() && $parsedRules) {
  $attrs = $this->getOption('attr') + $parsedRules;
  $this->setOption('attr', $attrs);
}
```

kristijanhusak:
@koichirose looks like it is. Can you just fix what i left in the comment, and i'll merge it in. Thanks!

koichirose:
Done! Thanks

kristijanhusak:
Thank you!

PR #383, opened 2017-09-19
[NEW][Features] Added support override data of ChoiceType

yemenifree:
sometimes maybe need to override choices the field with advanced logic, I know there is `query_builder` for `Entity` but its limit for a query & data from the database, but now can override data of all fields they base of `ChoiceType` field

Also, added simple test unit for this PR

example for using with `Entity` field

```php
        $this->add('languages', 'entity', [
            'class' => 'App\Language',
            'data_override' => function ($choices, $field) {
                $choices['testID’] = 'test';
                return $choices;
            }
        ]);
```


kristijanhusak:
Looks good, thanks!

PR #384, opened 2017-09-19
hidden function is added for Field.

bkayranci:
I think **hidden** function is required for fields.

### DIFF
--- a/src/Kris/LaravelFormBuilder/Fields/FormField.php
+++ b/src/Kris/LaravelFormBuilder/Fields/FormField.php
@@ -644,6 +644,18 @@ abstract class FormField
     }
 
     /**
+     * Hidden field.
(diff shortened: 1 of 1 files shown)


kristijanhusak:
Sorry but i cannot accept this. You can modify any field easily, and this is css stuff. If you need hidden value, just use input type hidden.

bkayranci:
@kristijanhusak However I have a using like following:

> $('#image').click(function(){
> $('#inputPhoto').click();
> });

because of that, I think we should have a hidden file input instead of hidden input.

kristijanhusak:
Just add the display none manually when creating a form. Different css doesnt mean different field type.

bkayranci:
OK. Thank you :+1:

PR #391, opened 2017-11-15
Fix: using properties that are populated via accessors

AkibaWolf:
Unable to use properties that are populated via accessor method in the 'property' option of Entity Type:

**In the Form class:**
`
->add('products', 'entity', [ ...
    'property' => 'name_translated',
... ])
`

**In the Product class:**
`public function getNameTranslatedAttribute()
{
    return ...;
}
`


kristijanhusak:
This solution should work for you:
https://github.com/kristijanhusak/laravel-form-builder/pull/369#issuecomment-427955679

PR #406, opened 2018-02-16
Allow for form level configuration

beghelli:
Hi Kris, thanks for this project. Very helpful. :)

### DIFF
--- a/src/Kris/LaravelFormBuilder/Fields/FormField.php
+++ b/src/Kris/LaravelFormBuilder/Fields/FormField.php
@@ -287,7 +287,7 @@ abstract class FormField
 
         if ($this->getOption('required') === true || isset($parsedRules['required'])) {
             $lblClass = $this->getOption('label_attr.class', '');
-            $requiredClass = $helper->getConfig('defaults.required_class', 'required');
+            $requiredClass = $this->getConfig('defaults.required_class', 'required');
 
             if (! str_contains($lblClass, $requiredClass)) {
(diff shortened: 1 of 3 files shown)


kristijanhusak:
@beghelli how you want to add a custom configuration for a form? As a property (like you did in a test), or in some other way?

I think this PR is a bit overcomplicated for such feature. We could have a `getConfig` method on a Form class, that would merge the custom configuration with the configuration read from the form helper class.

That way we could avoid injecting the custom config into the helper and handle everything on the form class.

kristijanhusak:
You can update fields to pull the config from the parent form that is injected through constructor. 
Basically my idea is to have the config property on each class, like you added, and to have a `getConfig` on the Form class, that will merge the default config from FormHelper and the config property that you set.
After that, update all calls to formHelper->getConfig to call Form->getConfig instead.

Not sure if i explained it well, but this solution does not require doing anything in the FormBuilder and FormHelper class, and avoids those conditional form helper injections.

beghelli:
Sure, I will give it a shot and update the pull request. Thanks!

beghelli:
@kristijanhusak updated the pull request with your suggestions. I did one change into the FormHelper class though, but just to avoid duplicating the logic of getting the configuration. Let me know what you think, thanks.

kristijanhusak:
@beghelli also please update all the changes to use spaces instead of tabs so it can be consistent.

kristijanhusak:
Thanks!

PR #407, opened 2018-02-18
Escape static field value

beghelli:
I didn't use Laravel's helper method e() to escape because I sense the templates were built avoiding any tight dependence with Laravel, so I tried to keep it that way.

### DIFF
--- a/src/views/static.php
+++ b/src/views/static.php
@@ -9,7 +9,7 @@
 <?php endif; ?>
 
 <?php if ($showField): ?>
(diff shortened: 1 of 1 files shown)


kristijanhusak:
Thanks!

sgrund:
How is it now possible to insert real html code into the form? Since this change we have many trouble with our forms. Maybe you should add a "raw" field type.

kristijanhusak:
@sgrund we could add a 'raw' => true option on the static field.

PR #411, opened 2018-02-23
Add field error class

n7olkachev:


kristijanhusak:
@n7olkachev can you explain a bit what is this PR for?

n7olkachev:
My bad, sorry.
Yep, it's useful because of bootstrap 4 validation, where you need to add `is-invalid` class on an `input` itself.

kristijanhusak:
Please update the default config to include that field class, set it to be empty string by default.

n7olkachev:
@kristijanhusak done

kristijanhusak:
Thanks!

PR #412, opened 2018-02-26
fixed error with collection type

### LINKED ISSUE #313: Collection only gets first item in request array and ignores the rest on form post
It seems that when I POST a form that has collections with a type of form, it only gets the first set in the array. Instead of using the built-in prototype (since that only seems to work for manually laying out forms), I use [jquery.repeater](https://github.com/DubFriend/jquery.repeater). I also dynamically generate my forms from a JSON string and a single Form class.

The request itself does have the additional rows, but laravel-form-builder only grabs the first line. What sort of options are supposed to be set in order to allow more than just one item?

kiperz:
Checking trueness on empty array returns false, so while creating form, request data is ommited. Fixes #313


kristijanhusak:
Thanks!

Issue #413, opened 2018-02-27
Multiple form in one page (More errorBag)


Fellner96:
Hi.. My problem is, i have multiple form in one simple page. I override the redirectIfNotValid method 8. line from `$response = $response->withErrors($this->getErrors())->withInput();` to `$response = $response->withErrors($this->getErrors(), $bag)->withInput();`. It's working, put's the errors to the custom bag. But when the page is refresh with error bag the fields doesn't show any errors... :/

kristijanhusak:
You could maybe structure your forms to be namespaced (have a name), so when you do anything, you can have everything in the same bag. I also think that would be easier to handle on the post action.

Fellner96:
`protected function addErrorClass()
    {
        $errors = $this->parent->getRequest()->session()->get('errors');

        if ($errors && $errors->has($this->getNameKey())) {
            $errorClass = $this->formHelper->getConfig('defaults.wrapper_error_class');
            $wrapperClass = $this->getOption('wrapper.class');

            if ($this->getOption('wrapper') && !str_contains($wrapperClass, $errorClass)) {
                $wrapperClass .= ' ' . $errorClass;
                $this->setOption('wrapper.class', $wrapperClass);
            }
        }
    }`

This always use the default error bag!!! And i can't override it becouse every field type extends this one...

Fellner96:
I will make a pull request, i think you dont understand my problem.

fredbradley:
Sorry to common on an issue that's been closed for a year, but I could do with a little support. 

I've tried to use this technique, in my `ContactForm` I've added `$errorBag = 'contactForm'`, but when I view the error bags after a failed validation, it still says 'default'!?

fredbradley:
Ignore me... fixed it - had something extra in our code that was complicating things.

Fellner96:
Working perfect for me.

tahafilali:
how can I handle this error ?
Thank you in advance !!

PR #414, opened 2018-02-27
#413 Multiple form in one page (More errorBag)

Fellner96:


kristijanhusak:
Thanks.

HoffmannAndras:
👏

hectnandez:
Hi,

In my application I have a several forms, but when I get the errors validation in one of them, an make a redirect back, the errors it's show it in all of them..

image

I just uptated to the version 1.6.50, but I'm still have this bug, do I miss something?

Thanks for the help.

rudiedirkx:
I think you have to explicitly name the error bag in every distinct form:

    /**
     * Define the error bag name for the form.
     *
     * @var string
     */
    protected $errorBag = 'default';

Only for forms that are on pages with other forms.

I think.

PR #419, opened 2018-03-22
Fixed issue 374; Error with Model and Name Form options

Tchilly:


kristijanhusak:
Added it in commit https://github.com/kristijanhusak/laravel-form-builder/commit/f7e1fbaba14276dd2634a85074888424a6209263 , with some additional checks to not break backward compatibility and tests. Thanks anyway!

Issue #270, opened 2016-08-14
Allow forms to alter their field values


rudiedirkx:
This is a reminder for me. My `getFieldValues`, before it was a native thing, allows child forms to change their values, so 0 checkboxes can become an empty array, or an empty string becomes null.

The new `getFieldValues` should have something like that too. There's at least 2 ways:

* get values per form, which means get values per field, and let every form alter & return them
* collect all values at once (currently) and then dissect them and pass them off to child forms

My `getFieldValues` also has `$nullables` like Laravel's `Model`, but that needs this.

rudiedirkx:
Relevant PR: https://github.com/kristijanhusak/laravel-form-builder/pull/272

kristijanhusak:
This is added in version 1.9. Closing. Thanks for help!

philyboysmith:
Could you give some guidance on how this works. I was hoping something like this would work for a custom field type, but alas not:

```
 public function alterFieldValues($value)
    {
        return myFunction($value)
    }
```

rudiedirkx:
The form is responsible for the alter, not the field. The only field that knows about this, is the child form, which proxies it to the Form class.

Fields don't have return values. That would be nice for v2.

rudiedirkx:
Also, there's no return value. See https://github.com/kristijanhusak/laravel-form-builder/blob/54b1af7fb95ca8e88a91afeb066e1102ff21728e/src/Kris/LaravelFormBuilder/Form.php#L1211

PR #422, opened 2018-04-04
Allow fields to transform values

Grafikart:
Fields can now implement a method `transformValue($value)` to transform values

Evolution of https://github.com/kristijanhusak/laravel-form-builder/issues/270


kristijanhusak:
@Grafikart let's then leave this to the developer to decide if he wants to add. Adding it to core doesn't add any value to the project because none of the core field types have it.

PR #424, opened 2018-04-22
Get namespace from composer autoload

d3v2a:


kristijanhusak:
What about PSR-0?

icfr:
Laravel app generally use psr-4 ;)

PR #435, opened 2018-06-06
Fixed bug where a custom closure would be interpreted as a string when using HTML 5 validation rules

yarbsemaj:


kristijanhusak:
Thanks!

Issue #416, opened 2018-03-02
checkbox help_block


warlord0:
When using a checkbox with a help_block I'm seeing the label come after the help_block.

```php
          ->add('ContractManagement', 'checkbox', [
            'label' => 'Contract Management',
            'wrapper' => ['class' => 'col-12 col-md-6 mb-2 mt-4'],
            'help_block' => [
              'text' => 'Select this if you wish to receive reminder notifications',
              'tag' => 'small',
              'attr' => [
                'class' => 'form-text text-muted'
              ]
            ]
          ])
```

It renders this:
```html

<div class="col-12 col-md-6 mb-2 mt-4"  >
   <input id="ContractManagement" name="ContractManagement" type="checkbox" value="1">
   <small class="form-text text-muted" >
      Select this if you wish to receive reminder notifications
   </small>   
   <label for="ContractManagement" class="control-label">Contract Management</label>
 </div>
```
![image](https://user-images.githubusercontent.com/9975039/36907750-4702445c-1e31-11e8-83a5-31a8882db9fd.png)

Is this a feature of the form-builder?

kristijanhusak:
It's a bug, it should be after the label. Same thing is for radio button. I'll fix it, or if you want, you can fix it and create PR.

PR #440, opened 2018-07-04
Fix: lable and help_block position bug with radio and checkbox

sagarnasit:
#416 Move label position before help_block for checkbox and radio button.


kristijanhusak:
Thanks!

PR #455, opened 2018-09-19
add const to all types

tresa02:
Hello Kris
I create const in interface for Form class to easy to use fields and faster to found.
When I use this package I should check the uses field every time.
when we use const we don't need search any time for special field

this form is readable than origin Form Class:

`class SongForm extends Form {
 public function buildForm() {
 $this ->add('name', Form::FIELD_TEXT) 
->add('lyrics', Form::FIELD_TEXTAREA) 
->add('publish', Form::FIELD_CHECKBOX); }
 }`

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,19 +83,20 @@ Form is created in path `app/Forms/SongForm.php` with content:
 namespace App\Forms;
 
 use Kris\LaravelFormBuilder\Form;
+use Kris\LaravelFormBuilder\Field;
 
 class SongForm extends Form
 {
     public function buildForm()
     {
         $this
-            ->add('name', 'text', [
+            ->add('name', Field::TEXT, [
                 'rules' => 'required|min:5'
             ])
-            ->add('lyrics', 'textarea', [
+            ->add('lyrics', Field::TEXTAREA, [
                 'rules' => 'max:5000'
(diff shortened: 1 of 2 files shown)


kristijanhusak:
I'd rather put this into a trait instead of an interface. Interface's purpose is not for this.

tresa02:
thank you kris but trait can not have constants .if we should not use interface we must define constants in Form Class so the good idea is Form class must extend to another class that have this constants if you agree with me i send pull request agian.
thank you.

tresa02:
thank you kris but trait can not have constants .if we should not use interface we must define constants in Form Class if you agree with me i send pull request agian.
thank you.

kristijanhusak:
@tresa02 oh, i didn't know traits cannot have constants, never tried it.
I don't like the idea of putting it in the form class. We could have it as a separate class, something like this:

```php
class Field {
    const TEXT = 'text';
    const TEXTAREA = 'textarea';
}
```

and then just use it:
```php
$this->add(Field::TEXT)->add(Field::TEXTAREA)
```

tresa02:
i fixed it 
tnx kris.

kristijanhusak:
Thanks!

PR #458, opened 2018-10-08
 Fix custom options with default options merge

MikeMnD:
Otherwise the default options overwrite the custom ones

    $this->options = $this->formHelper->mergeOptions($defaults, $this->options);

### DIFF
--- a/src/Kris/LaravelFormBuilder/Fields/FormField.php
+++ b/src/Kris/LaravelFormBuilder/Fields/FormField.php
@@ -563,7 +563,7 @@ abstract class FormField
         $this->options = $this->prepareOptions($options);
 
         $defaults = $this->setDefaultClasses($options);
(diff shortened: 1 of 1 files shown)


kristijanhusak:
They shouldn't override anything, there's a check if something custom is provided in `setDefaultClasses` method.

MikeMnD:
```
        $this->options = $this->formHelper->mergeOptions($this->allDefaults(), $this->getDefaults());
        $this->options = $this->prepareOptions($options);
        $defaults = $this->setDefaultClasses($options);
        $this->options = $this->formHelper->mergeOptions($defaults, $this->options);
```

Well if that's the case than maybe here: 
`$defaults = $this->setDefaultClasses($options);`

should be:

`$defaults = $this->setDefaultClasses($this->options);`

PR #469, opened 2018-12-19
Modify README.md

tomothumb:
Example code was old.

These are not used.
- `use Kris\LaravelFormBuilder\Field`;
- Field::TEXT
- Field::TEXTAREA
- Field::CHECKBOX


kristijanhusak:
I released new version that has Field enum.

PR #479, opened 2019-01-31
Make sure subvalues is an array

rudiedirkx:
A subform with checkboxes, with none checked, submits no values. If the complete form's values are retrieved with `getFieldValues(false)` (no nulls), the subform is empty (`null`), but `alterFieldValues` always expects an array.

I think I made this originally =) Sorry! Apparently I'm also the only one who uses it =)

### DIFF
--- a/src/Kris/LaravelFormBuilder/FormHelper.php
+++ b/src/Kris/LaravelFormBuilder/FormHelper.php
@@ -343,7 +343,7 @@ class FormHelper
             if (method_exists($field, 'alterFieldValues')) {
                 $fullName = $this->transformToDotSyntax($name);
 
-                $subValues = Arr::get($values, $fullName);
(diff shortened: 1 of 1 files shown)


SanderMuller:
This change introduced a bug in my custom form field (toggle), which uses the alterFieldValues method to convert unchecked checkbox to a fallback value, but now suddenly always returns an array while I'm still expecting an int/string, e.g. when doing `$form->getFieldValues()`, which again results in a `Array to string conversion (SQL: update...`  error when I fill the form values into my model.

SanderMuller:
Any update? Had to revert version and won't be able to update until this is fixed

rudiedirkx:
`alterFieldValues()` only works with an array. How did your code work without array? What's your code? What's your problem? Do you have a solution?

PR #481, opened 2019-02-07
Enable raw text inside labels

GeoSot:
Enable raw text inside label (disable escaping)

Example:

```
$this->add('title', 'text', [
    'label'      => '<small class="wow">test</small>',
    'label_attr' => [
        'raw' => true
    ]
]);
```

### DIFF
--- a/src/Kris/LaravelFormBuilder/FormBuilderServiceProvider.php
+++ b/src/Kris/LaravelFormBuilder/FormBuilderServiceProvider.php
@@ -84,12 +84,13 @@ class FormBuilderServiceProvider extends ServiceProvider
         $form = $this->app['form'];
 
         $form->macro('customLabel', function($name, $value, $options = []) use ($form) {
+            $escape = !array_get($options, 'raw', true);
             if (isset($options['for']) && $for = $options['for']) {
                 unset($options['for']);
-                return $form->label($for, $value, $options);
+                return $form->label($for, $value, $options, $escape);
             }
 
-            return $form->label($name, $value, $options);
+            return $form->label($name, $value, $options, $escape);
         });
     }
 


rudiedirkx:
You can do this with Laravel's `HtmlString`:

    'label' => new HtmlString('<small class="wow">test</small>'),

rudiedirkx:
Actually, you need a `laravelcollective/html` patch for that, because they did html escaping wrong. https://github.com/LaravelCollective/html/pull/597 should work. Apply composer patches with [cweagans/composer-patches](https://github.com/cweagans/composer-patches/).

mikeerickson:
I am keen to adding this method to core so another package is not required.

GeoSot:
@rudiedirkx 
Unfortunately, HtmlString() doesn't work as you said.
So it's a bit mess here  :/

mikeerickson:
I will add a working solution to the next release 1.30

rudiedirkx:
Escaping/not escaping shouldn't be the form builder's responsibility. This is why the form builder uses `laravelcollective/html`. We shouldn't repeat their features in our code. Maybe @kristijanhusak has an opinion.

kristijanhusak:
If i understood correctly, escaping was recently added to laravelcollective. Before that wasn't happening (according to the reported issue that we have, can't find it via mobile). We should see how can we disable escaping in laravelcollective, and just pass that option down, with default to not escape (as it worked before).

mikeerickson:
@kristijanhusak I am going to have a closer look at this issue in the coming days.  I hear what you are saying, and yes that is the case in a nutshell.

rudiedirkx:
I'd like it to work like `StaticType` works. The template html encodes:

    <?= e($options['value']) ?>

unless it's an `HtmlString`, because that's how `e()` works. That's how labels and options/choices should work too. If it doesn't, we have to fix `laravelcollective/html`, or not use it, but don't create yet another method to **not** html encode a `$value`.

Issue #402, opened 2018-02-02
Does it validate valid options?


perifer:
Can Laravel Form builder give an error if an value is given that is not specified in e.g a selects choices, without manually adding an in-rule?

E.g with 

```
->add('language', 'select', [
    'choices' => ['en' => 'English', 'fr' => 'French'],
])
```
it would give an error if `language` was not `en` or `fr`.

kristijanhusak:
You can add `empty_value` like it's shown here http://kristijanhusak.github.io/laravel-form-builder/field/select.html and then just add at least required rule. Something like this should work:

```php
->add('language', 'select', [
    'choices' => ['en' => 'English', 'fr' => 'French'],
    'empty_value' => 'Select language',
    'rules' => 'required'
])

```

perifer:
Thanks for the quick reply! I tried this and from what I can tell it does validate that the a value is selected, but it _doesn't_ give an error if I manipulate one of the options, e.g in the post request the value for language is set to "evil_language". 

To try to give a better example: A form where a user can upload a photo, and optionally tag friends in the photo by selecting from a select with friends `user_id`s. It's important that the user id submitted really is a friend of the user.

There are lots of places where you could eliminate validating that options are valid if this could be done for you. And it's not given that the input is required.

I didn't do any extensive research and it was a long time ago that I used the Symfony's form component – but [this old article](http://www.epixa.com/2011/08/create-and-validate-a-choice-list-in-a-symfony-2-form.html) (from 2011) seems to indicate that it does something like what I'm looking for:

> ... and even validates your input to make sure it is a valid option

kristijanhusak:
I don't think you'll be able to do this without using laravel's validator `in`. https://laravel.com/docs/5.5/validation#rule-in

If you have the choices, you know which are required, so you can do it the similar way like in the documentation:

```php
$choices = ['en' => 'English', 'fr' => 'French' ];
->add('language', 'select', [
    'choices' => $choices,
    'empty_value' => 'Select language',
    'rules' => [
        'required',
        Rule::in(array_keys($choices)
    ]
])
```

perifer:
What would you think of adding a more automatic way of doing this, and a way to opt in to that globally? So if you opted in to this, your forms choices would by default be validated.

perifer:
Ok, I’ll look into that. Thanks for your time!

perifer:
[...]

PR #491, opened 2019-02-23
Better field rules collection, with AfterCollectingFieldRules event

rudiedirkx:
My apps have custom fields and custom validators, but I want to use the form builder as pure as possible. An event for altering field rules would be perfect. Event `BeforeFormValidation` is too late, because it has prepared the entire validator already.

Patch is almost completely backward compatible, but not quite. The return type of `FormField::getValidationRules()` has changed from `array` to `Rules`. For new fields that's fine (because `Rules::fromArray`), but if a developer extends an existing field and expects an array from `parent::getValidationRules()`, that will crash.

### DIFF
--- a/src/Kris/LaravelFormBuilder/Events/AfterCollectingFieldRules.php
+++ b/src/Kris/LaravelFormBuilder/Events/AfterCollectingFieldRules.php
@@ -1,0 +1,54 @@
+<?php
+
+namespace Kris\LaravelFormBuilder\Events;
+
+use Kris\LaravelFormBuilder\Fields\FormField;
+use Kris\LaravelFormBuilder\Form;
+use Kris\LaravelFormBuilder\Rules;
+
+class AfterCollectingFieldRules
+{
+    /**
+     * The field instance.
+     *
+     * @var FormField
+     */
+    protected $field;
+
+    /**
+     * The field's rules.
+     *
+     * @var Rules
+     */
+    protected $rules;
+
+    /**
+     * Create a new after field creation instance.
+     *
+     * @param Form $form
+     * @param FormField $field
+     * @return void
+     */
+    public function __construct(FormField $field, Rules $rules) {
+        $this->field = $field;
+        $this->rules = $rules;
+    }
+
+    /**
+     * Return the event's field.
+     *
+     * @return FormField
+     */
+    public function getField() {
+        return $this->field;
+    }
+
+    /**
+     * Return the event's field's rules.
+     *
+     * @return Rules
+     */
+    public function getRules() {
+        return $this->rules;
+    }
+}
--- a/src/Kris/LaravelFormBuilder/Fields/FormField.php
+++ b/src/Kris/LaravelFormBuilder/Fields/FormField.php
@@ -7,6 +7,7 @@ use Kris\LaravelFormBuilder\Filters\FilterInterface;
 use Kris\LaravelFormBuilder\Filters\FilterResolver;
 use Kris\LaravelFormBuilder\Form;
 use Kris\LaravelFormBuilder\FormHelper;
+use Kris\LaravelFormBuilder\Rules;
 
 /**
  * Class FormField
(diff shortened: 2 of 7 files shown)


rudiedirkx:
I'm not a fan of `fireEvent()`, but I don't know how else to call the event dispatcher. `FormHelper` has a (protected) member `$formBuilder`, but it's always `null`, never set, never used.

kristijanhusak:
@rudiedirkx looks good. Would maybe implementing ArrayAccess solve the backward compatibity issue?

rudiedirkx:
Only for very simple access like `$x = $rules['rules']`. I wouldn't encourage `$rules['rules']['name'][] = 'min:10'` etc. By keeping `Rules` opaque, it can become more useful in the future. I think this small bc break (chance) is acceptable.

Maybe add some kind of deprecation notice where it encounters an `array` instead of a `Rules`? I don't know how that would work...

rudiedirkx:
Another use for this new event is #402. A developer can add an `in:...` rule to every `ChoiceType` automatically. I know I'm gonna.

kristijanhusak:
Ok, let's leave it like this. Thanks for PR.

rudiedirkx:
This is pretty bad code... I blame you too Kris!

To add a single `max:255` rule to a field if it doesn't have a `max` rule yet:

```
$events->listen(AfterCollectingFieldRules::class, function(AfterCollectingFieldRules $event) {
  $field = $event->getField();
  $ruler = $event->getRules(); // instanceof Rules

  $name = $field->getNameKey();
  $rules = $ruler->getRules()[$name] ?? []; // array, keyed by field names

  foreach ($rules as $rule) {
    if (strpos($rule, 'max:') === 0) {
      return;
    }
  }

  // To add it, whaaaat:
  $ruler->append(['rules' => [$name => [count($rules) => 'max:255']]]);
});
```

I have to explicitly add the `$name`, AND the rule's position in the field rules!?!? Because `append()` uses `array_replace_recursive()` which doesn't smart-merge numeric arrays like `array_merge`.

Slight more readable:

```
  $rules[] = 'max:255';
  $ruler->append(['rules' => [$name => $rules]]);
````

but still way too explicit. So I'm adding a few helper methods to `Rules`, which is why it exists:

```
$ruler->getFieldRules($fieldname = null);
$ruler->addFieldRule($rule, $fieldname = null);
$rules->setFieldRules(array $rules, $fieldname = null);
```

And `$fieldName` can be optional by remembering the field/origin when creating it (in `FormHelper->getFieldValidationRules()`.

I should beta test more next time, before PR'ing.

rudiedirkx:
Huh...

> remote: error: GH006: Protected branch update failed for refs/heads/master.
> remote: error: Required status check "continuous-integration/travis-ci" is expected.

I can't push to master? How do I make changes? I thought the Collaborator point was to not need PRs for everything..? I've pushed to a separate branch: https://github.com/kristijanhusak/laravel-form-builder/pull/new/better-rules-collection-methods

kristijanhusak:
@rudiedirkx try again, i removed restriction.

rudiedirkx:
Maybe that was on purpose. I don't know your desired flow.

[Push worked!](https://github.com/kristijanhusak/laravel-form-builder/commit/5a3a1c539cdd0e1a534882b46ea4c1d47ae4f4b2), and I removed the temp branch. I'll still make PRs for significant features, but this seemed small enough.

Agree? Disagree? @mikeerickson too?

mikeerickson:
@rudiedirkx I was going to approve this PR when I got home, but it has already been done.

rudiedirkx:
Yes, this PR is approved a while ago. But it was incomplete, so I wanted to complete the feature separately, without PR, quickly.

I don't know what the best flow is for a multi dev project. **Every** commit can't be a PR. But sometimes a review is a good thing.

mikeerickson:
@rudiedirkx Well, if we need to have a second set of eyes, just perform the typical PR request a PR review? Am i missing something?

kristijanhusak:
@rudiedirkx @mikeerickson lets always create PRs and have someone review it.
Push to master only as an emergency and tag releases.

mikeerickson:
Sounds good to me, I actually like it this way better. I like the workflow that requires peer code review, if for anything to catch even little issues.

PR #495, opened 2019-03-02
don't always automatically copy parent setOption(s) to children

rudiedirkx:
I use `AfterFieldCreation` to set options and add classes to some fields automatically. But if I `setOption` a `ChildFormType`, the option is copied to all children. For some options that's good, but for some it's not (field type class).

With this new option, the developer can choose not to copy options:

    $events->listen(AfterFieldCreation::class, function(AfterFieldCreation $event) {
      $event->getField()->setOption('copy_options_to_children', FALSE);
      $event->getField()->setOption('wrapper.class', 'foo');
    });

The default is `true`, so it's fully bc.


rudiedirkx:
Actually, I don't know any scenario where copying options would be good, but I'm assuming they exist...

PR #498, opened 2019-03-05
Laravel 5.8 Deprecates Array Helpers

GeoSot:


rudiedirkx:
Laravel is so lovely backward compatible. Not.

I see more Laravel `array_` functions. Which will be removed and which not? Maybe replace all of them?

mikeerickson:
> I see more Laravel array_ functions. Which will be removed and which not? Maybe replace all of them?

I have made some adjustments locally and will make sure I get them all adjusted

PR #376, opened 2017-08-27
[NEW][Features] Added support for field filters and request input values mutating.

unckleg:
I've implemented something similary to Zend form filters, so now developers can filter data that is passed through Form.

_Added:_

**Kris\LaravelFormBuilder\Fields\FormField:**
1. Property:
   - filtersOverride
1. Methods:
    -  setFilters(array $filters)
    -  getFilters()
    -  addFilter($filter)
    -  removeFilter($name)
    -  removeFIlters(array $filterNames)
    -  clearFilters()

**Kris\Form:**
1. Property:
    - lockFiltering
2. Methods:
    - filterFields
    - getFilters
    - lockFiltering
    - isFilteringLocked

**Kris\LaravelFormBuilder\Events\AfterFormCreation**
1. Methods:
    - filterFields

So now you can bind filters/filter to form field like this:
```
class SongForm extends Form
{
    public function buildForm()
    {
        $stringTrim = new StringTrim();
        $stripTags  = new StripTags();

        $this
            ->add('name', 'text', [
                'rules' => 'required',
                'filters' => [
                    $stringTrim,
                    $stripTags
                   // Or by Filter aliases/names
                   'StringTrim',
                   'StripTags'
                ],
                // This is used to allow override already assigned filters with same alias/name.
                'filters_override' => true,
            ])
        ;

        $this->getField('name')->removeFilters([
            'StringTrim',
            'StripTags'
        ]);
    }
}
```

**Modify fields:**
```
$this
    ->getField('name')
    ->addFilter(new Integer())
    ->removeFilters([
        'StringTrim',
        'StripTags',
        $filterObject
    ])
    ->removeFilter('StringTrim')
    ->setFilters([
        new SomeFilter(),
        new SecondFilter(),
        new FilterFromPackageCollection()
    ])
;
```

And if you e.g `StringToUpper` filter for field name and you passed in post this value: "some string"
the value you'll get in request is mutated. "SOME STRING".

PR #512, opened 2019-03-29
fixes #448: filters form subform fields

### LINKED ISSUE #448: Filters not working for child form?
After a bit of testing it seems that a child forms' filters are not getting applied. Only the parent form seems to be returning filtered input (checking values via this method `$form->getFieldValues()` and also inspecting `$request->all()`. And thoughts on what I may be missing when trying to retrieve a child forms' filtered data?

rudiedirkx:
Original issue: #448 by @ryanscherler

I don't like at all how filters change the global request data, but that's how they work. This fixes the scope problems. Fields in subforms didn't work. Fields like `address[street]` didn't work. This fixes both, but I don't like it at all. I'd love to put the filter logic in `alterFieldValues()` (so only through `getFieldValues()`), but that would break @unckleg's code (#376).

### DIFF
--- a/src/Kris/LaravelFormBuilder/Form.php
+++ b/src/Kris/LaravelFormBuilder/Form.php
@@ -1347,22 +1347,32 @@ class Form
     {
         // If filtering is unlocked/allowed we can start with filtering process.
         if (!$this->isFilteringLocked()) {
-            // Init required vars.
-            $filters = $this->getFilters();
-            $request = $this->getRequest();
+            $filters = array_filter($this->getFilters());
 
-            if (!empty($filters)) {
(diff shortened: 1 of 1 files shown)


kristijanhusak:
@rudiedirkx i don't think it's that bad, but we could clean these things up a bit. Maybe even move whole filtering code to separate class and do it there. I don't think it should all be in a Form class.

rudiedirkx:
If you're okay with the fix, let's merge it, and do refactoring/improvements in a new issue. There are more actual bugs to fix.

kristijanhusak:
Ok, merge it and put that refactoring somewhere on the roadmap.

PR #514, opened 2019-04-03
Pass form instance to "entity" field queryBuilder

cweiske:
This makes it possible to react to the current model values in
the query builder.

Example:
A "category" dropdown of type "entity".
Its queryBuilder adjusts the query to only show public categories.
When editing a model that already has a non-public category selected,
the query builder is now able to take this into account and
show public categories plus the currently selected non-public/internal
category.

### DIFF
--- a/src/Kris/LaravelFormBuilder/Fields/EntityType.php
+++ b/src/Kris/LaravelFormBuilder/Fields/EntityType.php
@@ -52,7 +52,7 @@ class EntityType extends ChoiceType
         }
 
         if ($queryBuilder instanceof \Closure) {
-            $data = $queryBuilder($entity);
+            $data = $queryBuilder($entity, $this->parent);
         } else {
             $data = $entity;
         }


kristijanhusak:
Looks good, thanks!

rudiedirkx:
The anonymous function can access `$this` (form) anyway, right? Why pass it in?

Do I understand correctly?:

```
class MyForm extends Form {
  function buildForm() {
    $this->add('category', 'entity', [
      'class' => Category::class,
      'query_builder' => function($model, $form) {
        // $form === $this, so why $form ?
      }
    ]);
  }
}
```

kristijanhusak:
@rudiedirkx yeah, i think you're right. Only potential difference is when the entity field is in nested child form, then parent doens't point to the main form.

rudiedirkx:
You want the smallest/deepest possible element, so you can work back/up from that to the main form, right? If the passed `$form` isn't the child form, there's no way to get the child form. It should be the smallest/deepest (child) form, not the parent form. Or even the deepest field, because you can work your way back/up from that to the child form > 'form' field > main form.

You can only add a 2nd argument once, so be sure it's useful.

cweiske:
In my case, I cannot access $this in the closure because I use a static function that is shared across many form elements, with `\Closure::fromCallback(['SharedClass', 'entityQueryBuilder'])`. I did not check if I could use `$this` of the closure itself, sorry.

Maybe passing the current value as second parameter would also suffice?

rudiedirkx:
Waat I didn't even know `Closure::fromCallable()`. I see your point. No `$this`. If @kristijanhusak thinks `parent` is the most useful, that's the one. I'll sneak on outahere now... =)

cweiske:
@kristijanhusak: For my use case, the current value would suffice.
Also, passing the `parent` removes all information about the current scope (name, value) and works only if the queryBuilder method knows which field is is used on.

So @rudiedirkx is correct in that passing the parent is not the best option.

What would you prefer: Passing the current value or the element itself as 2nd parameter to the queryBuilder closure?

PR #518, opened 2019-04-11
Fix #517 and unit test

### LINKED ISSUE #517: getFieldValues return only first item of form collection
When I use getFieldValues() to obtain the values of a form collection, it returns only the first collection item.

In my case, I have an EntriesForm class that contain a collection field (type form) that references the EntryForm class. When i submit EntriesForm with many values, getFieldValues() return only the first collection item.. However request()->all() return correctly the collection items.

**Result of getFieldValues():**
```php
array:1 [
  "entries" => array:1 [
    0 => array:1 [
      "name" => "test"
    ]
  ]
]
```

**Result of request()->all():**
```php
array:1 [
  "entries" => array:2 [
    0 => array:1 [
      "name" => "test"
    ]
    1 => array:1 [
      "name" => "test2"
    ]
  ]
]
```

**Reproduce:**
```php
class EntriesForm extends Form
{
    public function buildForm()
    {
        parent::buildForm();
        $this
            ->add('entries', 'collection', [
                'type' => 'form',
                'options' => [
                    'class' => EntryForm::class,
                    'label' => false
                ]
            ])
            ->add('submit', 'submit', ['label' => 'Save']);
    }
}
```

```php
class EntryForm extends Form
{
    public function buildForm()
    {
        parent::buildForm();
        $this
            ->add('name', Field::TEXT, [
                'rules' => 'required'
            ]);
    }
}
```

```php
class RegisterController extends Controller
{
    public function index(FormBuilder $formBuilder)
    {
        $form = $formBuilder->create(EntriesForm::class, [
            'url' => route('store'),
            'method' => 'POST'
        ]);

        return view('index', compact('form'));
    }

    public function store(FormBuilder $formBuilder)
    {
        $form = $formBuilder->create(EntriesForm::class);
        dump(request()->all());
        dd($form->getFieldValues());
    }
}
```

JamesHemery:
I've made a fix for issue #517

Fix is very easy, before `data` index of `getDefaults()` in `CollectionType` was `null`, i've replaced by an empty array. To make sure that it will not break anything, I wrote a test and started all the tests.

### DIFF
--- a/src/Kris/LaravelFormBuilder/Fields/CollectionType.php
+++ b/src/Kris/LaravelFormBuilder/Fields/CollectionType.php
@@ -35,7 +35,7 @@ class CollectionType extends ParentType
             'type' => null,
             'options' => ['is_child' => true],
             'prototype' => true,
-            'data' => null,
+            'data' => [],
             'property' => 'id',
             'prototype_name' => '__NAME__',
             'empty_row' => true,
--- a/tests/Fields/CollectionTypeTest.php
+++ b/tests/Fields/CollectionTypeTest.php
@@ -139,6 +139,25 @@ namespace {
             }
         }
 
+        /** @test */
+        public function it_get_all_items_of_collection_with_child_form()
+        {
+            $excepted = [
+                'entries' => [
+                    0 => ['name' => 'test'],
+                    1 => ['name' => 'test2'],
+                ]
+            ];
+
+            $this->request['entries'] = $excepted['entries'];
+
+            $form = $this->formBuilder->create('\LaravelFormBuilderCollectionTypeTest\Forms\EntriesForm');
+
+            $actual = $form->getFieldValues();
+
+            $this->assertEquals($excepted, $actual);
+        }
+
         /**
          * @test
          * @expectedException \Exception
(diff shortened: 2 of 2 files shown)


JamesHemery:
Travis tests with PHP 5.6 has failed, but the reasons for the failure do not make reference to the changes made.

rudiedirkx:
`data=null` is on purpose. I've explained in #517. I think config is enough to fix your problem.

rudiedirkx:
I like the test though. But it could test for all collection scenario's, changing config `empty_row` and `prefer_input`:

- `empty_row=false, prefer_input=false` (default) - [build elements] are validated and accepted
- `empty_row=true, prefer_input=false` (default) - [build elements] or [build elements + 1] are validated and accepted
- `prefer_input=true` (default) - any/all elements are validated and accepted

JamesHemery:
@rudiedirkx So this PR is not good for the merge?

rudiedirkx:
Yes, it's not good. If #517 is fixable with config, there's nothing to fix with code. If there's still something wrong, I want to know how. Discussion in #517.

PR #519, opened 2019-04-12
Attempt to keep data intact during validation error

rudiedirkx:
Use case:

- a collection of subforms with models as data
- a subform that uses the input model to print a checkbox label (enabled/disabled) or fieldset legend or something
- the parent form is submitted, but fails validation, any reason, and the entire form is printed again with `$oldInput` as data
- the input data for the collection is not a list of models anymore, it's the raw POST input, remembered in the session
- the subform breaks, because it expects `$this->model` to be a model, not a raw input array

This is my very crude attempt to keep the data models intact, while using the remembered `$oldInput` for values. With an option, off by default, because it'll probably break something.

Reproducible code, maybe:

```
class MainForm {
  function buildForm() {
    $tests = Test::fromUser(\Auth::user()); // Eloquent collection of Test models

    $this->add('tests', 'collection', [
      'type' => 'form',
      'empty_row' => true,
      'prototype' => false,
      'attempt_keep_data' => true,
      'data' => $tests,
      'options' => [
        'label' => false,
        'class' => ChildForm::class,
      ],
    ]);
  }
}

class ChildForm {
  function buildForm() {
    $this->add('enabled', 'checkbox', [
      'label' => (string) $this->model, // NEEDS A REAL MODEL
    ]);

    $this->add('custom_label', 'text', [
      'label' => 'Custom label for this test',
    ]);
  }
}
```

Without this patch, `'label' => (string) $this->model` would break, because `array to string conversion`, because `$this->model` is the collection item's remembered POST input (e.g. `[enabled => 1, custom_label => 'Waaat']`) instead of the original data model (a `Test`).

### DIFF
--- a/src/Kris/LaravelFormBuilder/Fields/CollectionType.php
+++ b/src/Kris/LaravelFormBuilder/Fields/CollectionType.php
@@ -40,6 +40,7 @@ class CollectionType extends ParentType
             'prototype_name' => '__NAME__',
             'empty_row' => true,
             'prefer_input' => false,
+            'attempt_keep_data' => false,
         ];
     }
 
@@ -100,14 +101,14 @@ class CollectionType extends ParentType
             $data = $currentInput;
         }
 
+        if ($data instanceof Collection) {
+            $data = $data->all();
+        }
+
         // Needs to have more than 1 item because 1 is rendered by default.
         // This overrides current request in situations when validation fails.
         if ($oldInput && count($oldInput) > 1) {
-            $data = $oldInput;
-        }
-
-        if ($data instanceof Collection) {
-            $data = $data->all();
+            $data = $this->overrideDataWithInput($data, $oldInput);
         }
 
         $field = new $fieldType($this->name, $type, $this->parent, $this->getOption('options'));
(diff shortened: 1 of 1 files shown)


kristijanhusak:
There is generally an issue with models converting to array at some point. We should maybe re-think this completely. I think workaround here would be to have a properly on the nested model that will be a label, instead of converting it to string.Not sure tho.

rudiedirkx:
This one isn't quite finished. If prototype is enabled, and used, and `prefer_input` is true, the new children get an array model during validation error (because there's no original data) instead of a Model. Accept a truty Model in `prototype` or `prefer_input` or `attempt_keep_data`, like in 349f6d5e39a356cb972ccc8253044395aa577ab0? Might get a bit wild:

```
$this->add('costs', 'collection', [
	'label' => false,
	'type' => 'form',
	'empty_row' => new SiteCost(),

    // Which of these 3? Or reuse empty_row? What if there's no empty_row?
	'prototype' => new SiteCost(),
	'prefer_input' => new SiteCost(),
	'attempt_keep_data' => new SiteCost(),

	'options' => [
		'class' => SiteCostForm::class,
		'label' => false,
		'wrapper' => [
			'class' => 'collection-child',
		],
	],
	'data' => $this->model ? $this->model->costs->concat([new SiteCost()]) : [],
	'wrapper' => [
		'class_append' => 'site-costs',
	],
]);
```

rudiedirkx:
Will fix later. Some time.

PR #525, opened 2019-04-29
default config form method setting

jva91:
add default config setting for form method

### DIFF
--- a/src/Kris/LaravelFormBuilder/Form.php
+++ b/src/Kris/LaravelFormBuilder/Form.php
@@ -46,7 +46,6 @@ class Form
      * @var array
      */
     protected $formOptions = [
-        'method' => 'GET',
         'url' => null
     ];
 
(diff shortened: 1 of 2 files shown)


rudiedirkx:
This fixes only your exact use case. There are more form options that could have other defaults, e.g. `client_validation`. See `Form::setFormOptions()`. If we could extend all options from configurable defaults, we can solve all cases, present and future. Like from an assoc array in `defaults.form_options`. Get it?

jva91:
Can u give me a list of settings that it would be nice to be overwritten by config? I will try to make a change for that

rudiedirkx:
You don't need a list, was my point. Any form option could be a default, present and future. Let the user decide what they want as defaults, as long as it's a form option as understood by `Form::setFormOptions()`, which is almost anything. Don't rely on specific options. Give it everything the user decides is a default.

Config:

```
return [
    'defaults' => [
        // ...
        'form_options' => [
            'method' => 'post',
            'client_validation' => false,
            'my_custom_form_option_that_my_custom_forms_understand' => ...,
            'a_form_option_that_exists_in_a_future_version' => ...,
        ],
```

PR #526, opened 2019-04-30
Try to avoid converting model to array.

kristijanhusak:
This could potentially fix issue that is trying to be fixed with https://github.com/kristijanhusak/laravel-form-builder/pull/519. I know we previously had bug reports where people didn't expect something to be an array  when they pass down a Model instance, or an object.

I barely tested this, so i need some testers to see if this could work out. If you know a better way to add a namespace to a model without breaking any structure, let me know.

This could maybe cause issues with lots of nested data. Converting array to object does it only on top level.

@rudiedirkx it would be great if you could test this and see if it works.

### DIFF
--- a/src/Kris/LaravelFormBuilder/Fields/FormField.php
+++ b/src/Kris/LaravelFormBuilder/Fields/FormField.php
@@ -231,14 +231,7 @@ abstract class FormField
      */
     protected function getModelValueAttribute($model, $name)
     {
-        $transformedName = $this->transformKey($name);
-        if (is_string($model)) {
-            return $model;
-        } elseif (is_object($model)) {
-            return object_get($model, $transformedName);
-        } elseif (is_array($model)) {
-            return Arr::get($model, $transformedName);
-        }
+        return $this->formHelper->getModelValue($model, $name);
     }
 
     /**
--- a/src/Kris/LaravelFormBuilder/Form.php
+++ b/src/Kris/LaravelFormBuilder/Form.php
@@ -1052,14 +1052,18 @@ class Form
         }
 
         $dotName = $this->formHelper->transformToDotSyntax($this->getName());
-        $model = $this->formHelper->convertModelToArray($this->getModel());
+        $model = $this->getModel();
         $isCollectionFormModel = (bool) preg_match('/^.*\.\d+$/', $dotName);
         $isCollectionPrototype = strpos($dotName, '__NAME__') !== false;
 
-        if (!Arr::get($model, $dotName) && !$isCollectionFormModel && !$isCollectionPrototype) {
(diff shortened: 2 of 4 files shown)


rudiedirkx:
This solves very few cases... The problem is that `$this->model` is not the original model, not just that it's an array instead of object. Devs expect it to be the original, which makes sense.

A better solution IMO would be to never touch `$this->model` and put all the attribute-getter-logic in one location, to be called at the last instant, to do only that: get a value (from original model + post data + other sources/logic). But that would probably require a bc breaking refactor.

kristijanhusak:
@rudiedirkx initially that was added to allow adding namespace to a form. Example:

```php
FormBuilder::create(MyForm::class, [
  'name' => 'my_namespace',
  'model' => $model
]);
```

And then same code was later used to populate values for child forms.
It should solve this simple case in situations where `$model` is object or array.

kristijanhusak:
@rudiedirkx can you give me an example how would you do that logic without changing the model?

rudiedirkx:
Oef. My #519 fix is good enough for me for now, so I don't know if I want to spend time on my perfect fix. I'll try something soon, but not very soon.

It'd be something like a second source (like `->model` is the first). The first is always exactly what the dev assigns. The (optional) second could be post form data. Both would exist at the same time. The dev would use `->model`, predictably, and the form builder might use the second source to fill field values.

kristijanhusak:
@rudiedirkx ok, lets close this and merge your fix. It is a big change, so we will do it eventually.

rudiedirkx:
Man, I'm having this problem again, but slightly different. Parent form has no model, and adds a 'form' field with empty default. Because the child form is also used as only/parent form, it expects a real model, not an array. No collections anywhere.

Parent form:

```
$this->add('translation', 'form', [
  'class' => ContentTranslationForm::class,
  'data' => ['locale' => app()->getLocale()],
  'value' => new Page(),
]);
```

but that doesn't work, because `setupNamedModel()` breaks it, because the model (`new Page`) doesn't have a `translation` property. And why would it? I'm passing in the literal value. `translation` is a parent form field, not a child form field.

Not better, but different:

```
$this->add('translation', 'form', [
  'class' => ContentTranslationForm::class,
  'data' => ['locale' => app()->getLocale()],
  'value' => ['translation' => new Page()],
]);
```

Now the 'model' does have a `translation`, so the entire model is left alone, so now it's the array with the model inside. Still not the literal `value`.

The only one that works:

```
$this->add('translation', 'form', [
  'class' => ContentTranslationForm::class,
  'data' => ['locale' => app()->getLocale()],
  'value' => (new Page())->setRelation('translation', new Page(['admin_title' => ''])),
]);
```

but that's just too ridiculous... (The outside `Page` is used, because it has a non-empty `translation` property.)

Why is `setupNamedModel()` so weird?? It shouldn't use the parent form's child form field name (`translation` in this case), because it already knows the value...

PR #529, opened 2019-05-22
errors when session not set.

Fellner96:
now you can build your form in your queue if you need to calculate data with it


rudiedirkx:
Good enough for me. I am curious though: Why does a queue have a Request, but not a Session?

PR #551, opened 2019-09-06
Update phpunit and testbench to allow testing for laravel 6

migueldamaso:
This change replaces some methods and helpers that have a new sintax to use in the laravel 6. The test packages have also changed since the current version of the package orchestra/testbench does not test for laravel 6.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,9 +19,9 @@
     ],
     "require": {
         "php": ">=5.6.0",
-        "laravelcollective/html": "5.*",
-        "illuminate/database": "5.*@dev",
-        "illuminate/validation": "5.*@dev"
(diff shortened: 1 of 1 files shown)


mikeerickson:
As you have likely seen, these changes have broken TravisCI.  I won’t be able to review this until late this weekend.  If you want to review the issues and update accordingly, please do so.  Until the CI process is passing, this PR will not be merged.

migueldamaso:
Ok. To update to the orchestra/testbench that tests laravel 6 it's required to update phpunit to at least 8.0.0 which requires php 7.2. So i don't think i can update and pass the tests. what i may do is send a pull request with the actual fixes only

PR #567, opened 2019-11-26
Change ParentType so it merges arrays to prevent overwriting default options of children.

### LINKED ISSUE #445: Choice children attribute is overwritten
framwork: Laravel 5.6；
laravel-form-builder: 1.15.1

``` php
        $form->add('permissions','choice',[
            'choices' => $permission_select,
            'multiple' => true,
            'rules' => 'present|boolean',
            'expanded' => true,
            'label' => false,
            'wrapper' => ['class' => 'row'],
            'choice_options' => [
                'wrapper' => ['class' => 'col-sm-3'],
                'label_attr' => ['class' => 'label fix_line_height margin-r-5 bg-green'],
            ],
            'attr' => ['class' => ''],
            'selected' => $selected_permission,
        ]);
```
``` html
<div class="col-sm-3">
    
    <input class="" multiple="" pattern="0|1" title=" 必须为布尔值。" name="permissions[]" type="checkbox" value="admin.administratorManage.index">
    
    <label for="permissions_admin.administratorManage.index" class="label fix_line_height margin-r-5 bg-green">管理员-浏览</label>


        </div>
```
No default ID attribute，I tried to modify the code(class ParentType)：
``` php
    /**
     * @inheritdoc
         */
    public function setOption($name, $value)
    {
        parent::setOption($name, $value);

        foreach ((array) $this->children as $key => $child) {
//            $this->children[$key]->setOption($name, $value);
            $this->children[$key]->setOption($name, is_array($value) ? array_merge($value,$this->children[$key]->getOption($name)) : $value);
        }
    }
```
Now working properly.

Luukvdo:
This PR is meant to fix #445.

The application I'm working on does not display the checkboxes, but rather it displays a label. Clicking on the label changes the checkbox.
For this to work the label has the `for` attribute, which matches the checkbox `id` attribute.

Without this fix the id is not passed to the checkbox causing the checkbox to not work properly.

### DIFF
--- a/src/Kris/LaravelFormBuilder/Fields/ParentType.php
+++ b/src/Kris/LaravelFormBuilder/Fields/ParentType.php
@@ -103,7 +103,7 @@ abstract class ParentType extends FormField
 
         if ($this->options['copy_options_to_children']) {
             foreach ((array) $this->children as $key => $child) {
-                $this->children[$key]->setOption($name, $value);
+                $this->children[$key]->setOption($name, is_array($value) ? array_merge($value, $this->children[$key]->getOption($name)) : $value);
             }
         }
     }
--- a/tests/Fields/ChoiceTypeTest.php
+++ b/tests/Fields/ChoiceTypeTest.php
@@ -1,5 +1,6 @@
 <?php
 
+use Kris\LaravelFormBuilder\Fields\CheckableType;
 use Kris\LaravelFormBuilder\Fields\ChoiceType;
 
 class ChoiceTypeTest extends FormBuilderTestCase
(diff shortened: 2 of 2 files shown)


Luukvdo:
Just found a flaw where the required rule is applied on all checkboxes, so this is not ready to merge. I'll try to fix it though.

rudiedirkx:
That's why the `copy_options_to_children` option exists (default `true`). I don't know if it's in the latest release. And why I don't use `ChoiceType`.

Luukvdo:
Thanks for the tip @rudiedirkx, setting `copy_options_to_children` to false fixes it.
It is in the latest release but I didn't realise that's what I needed to fix my case.

I'll close this PR.

PR #577, opened 2020-01-10
Fix `mergeOptions` behavior

yaquawa:
`\Kris\LaravelFormBuilder\FormHelper::mergeOptions` merges options with `rules` option behavior incorrectly. It should replace the rules entirely or just appends new rules. 
The current strategy may cause the merged rules duplicated, and has no way to "replace" the default rules entirely.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
         }
     ],
     "require": {
-        "php": ">=7.0",
+        "php": ">=5.6.0",
         "laravelcollective/html": "5.*|^6.0",
         "illuminate/database": "5.*@dev|^6.0",
         "illuminate/validation": "5.*@dev|^6.0"
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
         }
     ],
     "require": {
-        "php": ">=5.6.0",
+        "php": ">=7.0",
         "laravelcollective/html": "5.*|^6.0",
         "illuminate/database": "5.*@dev|^6.0",
         "illuminate/validation": "5.*@dev|^6.0"
--- a/composer.json
+++ b/composer.json
@@ -18,14 +18,13 @@
         }
     ],
     "require": {
-        "php": ">=5.6.0",
-        "laravelcollective/html": "5.*|^6.0",
-        "illuminate/database": "5.*@dev|^6.0",
-        "illuminate/validation": "5.*@dev|^6.0"
+        "php": ">=7.1",
+        "laravelcollective/html": "5.6.*|^6.0",
+        "illuminate/database": "5.6.*@dev|^6.0",
+        "illuminate/validation": "5.6.*@dev|^6.0"
     },
     "require-dev": {
-        "phpunit/phpunit": "~5.0",
-        "orchestra/testbench": "~3.4"
+        "orchestra/testbench": "~3.6"
     },
     "extra": {
         "branch-alias": {
--- a/src/Kris/LaravelFormBuilder/Fields/FormField.php
+++ b/src/Kris/LaravelFormBuilder/Fields/FormField.php
@@ -257,16 +257,19 @@ abstract class FormField
      * Prepare options for rendering.
      *
      * @param array $options
-     * @return array
+     * @return array The parsed options
      */
     protected function prepareOptions(array $options = [])
     {
         $helper = $this->formHelper;
+
+        $this->options = $this->prepareRules($options);
+        $this->options = $helper->mergeOptions($this->options, $options);
+
         $rulesParser = $helper->createRulesParser($this);
         $rules = $this->getOption('rules');
         $parsedRules = $rules ? $rulesParser->parse($rules) : [];
 
-        $this->options = $helper->mergeOptions($this->options, $options);
 
         foreach (['attr', 'label_attr', 'wrapper'] as $appendable) {
             // Append values to the 'class' attribute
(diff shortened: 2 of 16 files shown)


kristijanhusak:
There are two issues with your changes:

* mergeOptions is just a helper, so we should not put feature specific things in it
* these hacks with `+` are completely not acceptable

I think you can set rules as an array, it should properly merge them, not replace them.

yaquawa:
@kristijanhusak 

The current implementation has several problems:

## can't replace the default rules

```php
$defaultRules = ['rule1', 'rule2'];
$myRules = ['rule3'];
$replaceRules = ture;

// => got ['rule3', 'rule2'], but it should be ['rule3']
mergeOptions($defaultRules, $myRules, $replaceRules);
```

## may cause the merged rules duplicated
```php
$defaultRules = ['rule1', 'rule2'];
$myRules = ['rule2'];


// => got ['rule2', 'rule2'], but it should be ['rule1','rule2']
mergeOptions($defaultRules, $myRules);

// To make it correct you have to do this, 
// which is not intuitive and not predictable if the default rules got changed.
$myRules = [1=> 'rule2'];
mergeOptions($defaultRules, $myRules);
```

yaquawa:
@kristijanhusak 
> mergeOptions is just a helper, so we should not put feature specific things in it

So should we move it to the Form class?

> these hacks with + are completely not acceptable

I'd like to say it's an "option" after all, it can be something else like "appended_rules". What do you think?

yaquawa:
The default options can be very handy when building custom field.
But the current implementation has several problems as I described before.
I just want make some enhancements to this package, any better ideas?

kristijanhusak:
Can you give me a real example how to cause this rules issue? At what moment do you override it and where?

yaquawa:
@kristijanhusak 
For example, if you created a custom form field `PasswordField`, which can add some additional feature, like checking if it's a good password by checking the complexity of it by setting some default rules. Users of it can add additional rules like `required` or `confirmed` etc.  Or replace the default rules if they want.

And above all, I think the current implementation can cause unexpected behavior because it's not intuitive and not predictable when merging rules.

kristijanhusak:
Lets add `rules_append` property that will append these to rules. instead od overwriting it. I think we already have something like this for other option.

yaquawa:
@kristijanhusak 
> I think we already have something like this for other option.

Sorry about that, I didn't notice that at all..!
I've just renamed the option name to `rules_append ` and reverted the overwriting behavior.

kristijanhusak:
@yaquawa please rebase this branch from master, i merged other PRs and this one have all of the changes.

yaquawa:
@kristijanhusak Thanks for the mergings!!

> please rebase this branch from master

Why should we rebase this branch onto the master..? I don't think we need a rebase here😯

kristijanhusak:
@yaquawa there are some changes from the previous PRs, around multiple remove method. Just rebase from the current master.

Also, let's not do anything in the FormHelper class regarding this. Add a method to a Form class, and call it in the `prepareOptions` where we do all these mappings for the options.

Also, consider that Rule can also be a class or a closure, so take care of that when normalizing.

yaquawa:
@kristijanhusak 
Added some additional fixes to support Laravel >=5.6.

As of Laravel 5.6, [closure rules](https://laravel.com/docs/5.7/validation#using-closures) was supported.

I think the old implementation should be updated to align to the Laravel's one now to avoid confusion for modern user of Laravel.

Also maybe it's the right time to upgrade to php 7 now.

rudiedirkx:
This seems very big and invasive. You guys okay if I test this properly when I'm back (in 2 weeks)? I'd like to see what it does in all my Laravel projects. If it doesn't break anything, at least it's bc.

kristijanhusak:
@rudiedirkx yes please. And also do a code review, i'm not completely following latest changes to Laravel.

yaquawa:
@rudiedirkx Yes, thanks :)

There are 2 breaking changes of this PR:
* "closure rules" adopted to Laravel's one
* Require Laravel >=5.6(>=PHP 7.1)

rudiedirkx:
Laravel's closure clashes with the form builder's closure support. They mean and do something else. With this patch, the form builder closure is replaced with Laravel's closure. That might break some projects that use the form builder's closure. (I've never used it and I don't know when/how/why it would be used, but some might.)

If we keep the form builder's closures, we can ALSO support Laravel's closures: a closure that returns a closure. If we remove the form builder's, it's broken. It's weird, but it would be fully bc. To support both and use Laravel's closure rule:

```
$closureRule = function($attribute) { // form builder closure
    return function($attribute, $value, $fail) { // Laravel closure
        if ($value != 'foo') {
            $fail("$attribute is wrong");
        }
    };
};
```

But that's weird =) I'm fine with this patch, because I don't use or understand the form builder's current closure feature (so to remove it), but it's up to @kristijanhusak because he probably made it and knows why.

All the rest is great. @kristijanhusak You should change Travis CI to test in PHP 7.1, 7.2 and 7.3, and remove 5.6 and 7.0, so the tests don't fail.

rudiedirkx:
I like this one. @kristijanhusak Are you okay with dropping closure support?

kristijanhusak:
@rudiedirkx if you think code is ok we can merge it in. I'll let you take care of that once you review it.

rudiedirkx:
Excellent. I've reviewed it and I like it. @kristijanhusak You're the only who can change Travis CI settings, so can you change the PHP versions to only test in `7.1`, `7.2`, `7.3`? Tests fail because this PR changes PHP version requirements.

rudiedirkx:
Ehm. What happened there, Github? Seems like merging went wrong or weird. I don't like this one. Let me retry.

rudiedirkx:
@kristijanhusak If you unprotect `master` for a second, I can rebase those 7 commits into one. Github doesn't accept my push:

```
To github.com:kristijanhusak/laravel-form-builder.git
 ! [remote rejected] master -> master (protected branch hook declined)
error: failed to push some refs to 'git@github.com:kristijanhusak/laravel-form-builder.git'
```

kristijanhusak:
@rudiedirkx it's ok, just leave them as it is.
@yaquawa lets figure out why tests are failing and fix them with the Form::group PR.

rudiedirkx:
Tests are failing because Travis CI PHP versions are wrong now. Only you can change that, in Travis CI.

rudiedirkx:
@kristijanhusak Oh, is that editable in the repo... =) I also added 7.3 and 7.4.

PR #578, opened 2020-01-11
Fix exception thrown if `formOptions` contains any non-string value.

yaquawa:
## Issue

When rendering the form with `$formOptions` contains any non-string value like

```php
class MyForm extends BaseForm
{
    protected $formOptions = [
        'foo' => ['bar'], // array value
    ];
}
```

Template:
```php
// This will throw an exception.
<?= Form::open($formOptions) ?>
```

## Solution
Move form attributes to `$formOptions['attr']`, only pass `$formOptions['attr']` and reserved options to the view.

After this PR:
```php
// in template the `foo` was removed from `$formOptions`, 
// so the following code won't throw exception.
<?= Form::open($formOptions) ?>

// but we can still access the original option 
// by calling `$form->getFormOption('foo')`
```

### DIFF
--- a/src/Kris/LaravelFormBuilder/Form.php
+++ b/src/Kris/LaravelFormBuilder/Form.php
@@ -47,7 +47,8 @@ class Form
      */
     protected $formOptions = [
         'method' => 'GET',
-        'url' => null
+        'url' => null,
+        'attr' => [],
     ];
 
     /**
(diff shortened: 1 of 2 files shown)


kristijanhusak:
Make sure to maintain backward compatibility.

yaquawa:
@kristijanhusak Done. Thanks for reminding me that.

Issue #572, opened 2020-01-04
How to create a row with multiple fields?


yaquawa:
Can I create a "[form row](https://getbootstrap.com/docs/4.4/components/forms/#form-row)" with this package?

Anyway, the following code works for me.
Maybe this type of field can be added in the build-in ones? 

```php
<?php namespace App\Forms\Fields;

use Kris\LaravelFormBuilder\Fields\ParentType;

class FieldsType extends ParentType
{
    protected function getTemplate()
    {
        return 'form.fields.fields';
    }

    /**
     * @return mixed|void
     * @throws \Exception
     */
    protected function createChildren()
    {
        $this->children = [];
        $fields = $this->getOption('fields');
        $requiredOptions = ['type', 'name'];

        foreach ($fields as $field) {
            foreach ($requiredOptions as $requiredOption) {
                if (empty($field[$requiredOption])) {
                    throw new \Exception("Fields field [{$this->name}] requires [{$requiredOption}] option");
                }
            }

            $type = $field['type'];
            $name = $field['name'];
            $options = $field['options'] ?? [];
            $fieldType = $this->formHelper->getFieldType($type);
            $parentName = $this->getParent()->getName();
            $field = new $fieldType($parentName ? "{$parentName}[{$name}]" : $name, $type, $this->parent, $options);
            $this->children[] = $field;
        }
    }

    /**
     * @inheritdoc
     */
    public function getAllAttributes()
    {
        return $this->formHelper->mergeAttributes($this->children);
    }

    /**
     * @inheritdoc
     */
    public function getValidationRules()
    {
        return $this->formHelper->mergeFieldsRules($this->children);
    }
}
```

`resources/views/form/fields/fields.blade.php`

```blade
@extends('form.fields.base-layout')

@section('field')
    <div class="form-row">
        @foreach((array)$options['children'] as $child)
            {!! $child->render() !!}
        @endforeach
    </div>
@overwrite
```

`resources/views/form/fields/base-layout.blade.php`

```blade
@if($showLabel || $showField)
    @if($options['wrapper'] !== false)
        <div {!! $options['wrapperAttrs'] !!}>
    @endif

    @if ($showLabel && $options['label'] !== false && $options['label_show'])
        {!! Form::customLabel($name, $options['label'], $options['label_attr']) !!}
    @endif

    @if ($showField)
        @yield('field')
        @include('form.help_block')
        @include('form.errors')
        @yield('field_bottom')
    @endif

[...]

PR #580, opened 2020-01-23
Add a `ContainerType` field

yaquawa:
This PR adds a `ContainerType` field which solves #572.
It was branched from #577 , so you might only want to see the diff at https://github.com/kristijanhusak/laravel-form-builder/commit/d016cba31cabfded082919f0bebaa8efa106bc06

`ContainerType` field acts as a "container", which means even it has children fields, but it'll make the hierarchy of it's children as same as itself. In other words, the field's attributes and rules are be replaced by its children's ones(similar to `ChildFormType`). With this, users can easily implement `fieldset` or similar things by just specifying a custom template.

## Usage
```php
<?php

namespace App\Forms;

class MyForm extends Form
{
    public function buildForm()
    {
        $this
            ->add('date', 'container', [
                'label' => 'DATE',
                'fields' => [
                    [
                        'type' => 'date',
                        'name' => 'start_date',
                        'options' => [
                            'label' => false,
                            'rules' => 'required|date|max:255',
                        ],
                    ],
                    [
                        'type' => 'date',
                        'name' => 'end_date',
                        'rules' => 'date|max:255',
                        'options' => [
                            'label' => false,
                            'rules' => 'date|max:255',
                        ],
                    ]
                ],
            ]);
    }
}
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ composer.lock
 coverage
 *.taskpaper
 NOTES.md
+.phpunit.result.cache
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
         }
     ],
     "require": {
-        "php": ">=7.0",
+        "php": ">=5.6.0",
         "laravelcollective/html": "5.*|^6.0",
         "illuminate/database": "5.*@dev|^6.0",
         "illuminate/validation": "5.*@dev|^6.0"
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
         }
     ],
     "require": {
-        "php": ">=5.6.0",
+        "php": ">=7.0",
         "laravelcollective/html": "5.*|^6.0",
         "illuminate/database": "5.*@dev|^6.0",
         "illuminate/validation": "5.*@dev|^6.0"
--- a/composer.json
+++ b/composer.json
@@ -18,14 +18,13 @@
         }
     ],
     "require": {
-        "php": ">=5.6.0",
-        "laravelcollective/html": "5.*|^6.0",
-        "illuminate/database": "5.*@dev|^6.0",
-        "illuminate/validation": "5.*@dev|^6.0"
+        "php": ">=7.1",
+        "laravelcollective/html": "5.6.*|^6.0",
+        "illuminate/database": "5.6.*@dev|^6.0",
+        "illuminate/validation": "5.6.*@dev|^6.0"
     },
     "require-dev": {
-        "phpunit/phpunit": "~5.0",
-        "orchestra/testbench": "~3.4"
+        "orchestra/testbench": "~3.6"
     },
     "extra": {
         "branch-alias": {
--- a/src/Kris/LaravelFormBuilder/Fields/ContainerType.php
+++ b/src/Kris/LaravelFormBuilder/Fields/ContainerType.php
@@ -1,0 +1,69 @@
+<?php
+
+namespace Kris\LaravelFormBuilder\Fields;
+
+class ContainerType extends ParentType
+{
+    protected function getTemplate()
+    {
+        return 'container';
+    }
+
+    /**
+     * @inheritdoc
+     */
+    protected function getDefaults()
+    {
+        return [
+            'fields' => [],
+            'container_class' => 'form-group',
+        ];
+    }
+
+    /**
+     * @return mixed|void
+     * @throws \InvalidArgumentException
+     */
+    protected function createChildren()
+    {
+        $this->children = [];
+        $fields = $this->getOption('fields');
+        $requiredOptions = ['type', 'name'];
+
+        foreach ($fields as $field) {
+            foreach ($requiredOptions as $requiredOption) {
+                if (empty($field[$requiredOption])) {
+                    throw new \InvalidArgumentException("Fields field [{$this->name}] requires [{$requiredOption}] option");
+                }
+            }
+
+            $parentName = $this->parent->getName();
+            $rawName = $field['name'];
+            $name = $parentName ? "{$parentName}[{$rawName}]" : $rawName;
+            $options = $this->formHelper->mergeOptions($field['options'] ?? [], ['attr' => ['id' => $name]]);
+            $type = $field['type'];
+            $fieldType = $this->formHelper->getFieldType($type);
+            $field = new $fieldType($name, $type, $this->parent, $options);
+            $value = $this->getModelValueAttribute($this->parent->getModel(), $rawName);
+            $field->setValue($value);
+            $this->children[$rawName] = $field;
+        }
+    }
+
+
+    /**
+     * @inheritdoc
+     */
+    public function getAllAttributes()
+    {
+        return $this->formHelper->mergeAttributes($this->children);
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function getValidationRules()
+    {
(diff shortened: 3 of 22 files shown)


kristijanhusak:
@yaquawa what is difference between using this container and having a child form? You can use custom template for child form and achieve same results. Or i'm missing something. Please elaborate.

yaquawa:
@kristijanhusak `ChildForm` will include the namespace of itself. This won't. So if use child form, the example shown above will get us input name: `date[start_date]` and `date[end_date]`. If you use `ContainerType` you'll get just `start_date` and `end_date`.

yaquawa:
Another way to do this is adding an option to the exciting `ChildForm` Class, for example maybe : `collapse_namespace=>true` to get rig of the parent's namespace.

kristijanhusak:
@yaquawa you could use `compose()` method on Form class.
Here's docs to it http://kristijanhusak.github.io/laravel-form-builder/form/methods-and-properties.html#compose and here's PR where it was introduced https://github.com/kristijanhusak/laravel-form-builder/pull/108

yaquawa:
@kristijanhusak if you use `compose` to combine form fields,  then how can you render the combined fields in a container?

yaquawa:
Should this be nicer for you (use `ChildFormType`)?

```php
<?php

namespace App\Forms;

class MyForm extends Form
{
    public function buildForm()
    {
        $this
            ->add('date', 'form', [
                'label' => 'DATE',
                'namespace' => false,
                'fields' => [
                    [
                        'type' => 'date',
                        'name' => 'start_date',
                        'options' => [
                            'label' => false,
                            'rules' => 'required|date|max:255',
                        ],
                    ],
                    [
                        'type' => 'date',
                        'name' => 'end_date',
                        'rules' => 'date|max:255',
                        'options' => [
                            'label' => false,
                            'rules' => 'date|max:255',
                        ],
                    ]
                ],
            ]);
    }
}

```

kristijanhusak:
You can use custom template. I don't like the idea of defining these fields array. If you need to define it like this, it definitely means it needs to go to separate class. Let's try to do something that will work in similar way as compose, but to allow setting up the container.

yaquawa:
> You can use custom template. I don't like the idea of defining these fields array. If you need to define it like this, it definitely means it needs to go to separate class. Let's try to do something that will work in similar way as compose, but to allow setting up the container.

I referred that before that you can't just simply use custom template because of the namespace.

Do you mean use `compose()` (which is the same thing as `add()`)  and render them by hand (`$field->render()`)?

yaquawa:
> Let's try to do something that will work in similar way as compose, but to allow setting up the container.

That what this PR wants to introduce. Am I do something wrong?😥

yaquawa:
@kristijanhusak 
How about something like this?

```php
<?php

namespace App\Forms;

abstract class FieldGroup
{
    abstract public function render();
}

class MyGroup extends FieldGroup
{
    public function render()
    {
        echo '<div class="container">';

        foreach ($this->fields as $field) {
            $field->render();
        }

        echo '</div>';
    }
}

class MyForm extends Form
{
    public function buildForm()
    {
        $this
            ->add('start_date', 'date', ['group' => MyGroup::class])
            ->add('end_date', 'date', ['group' => MyGroup::class]);
    }
}

```

kristijanhusak:
@yaquawa i meant to allow using child form for same purpose, but modify the html structure with some options. I'll let @rudiedirkx give his opinion on this.

yaquawa:
Thanks @kristijanhusak !

I change my code a little bit above to use a method called `group` to group fields together, so the code looks more self descriptive. inside the `group` method, just extract the fields user added, then add it to the form, as the same time add them as a group field to the `$this->fields_for_render` or something like that. At the same time, we can get more loose coupling by separating the `$this->fields` and `$this->fields_for_render`.

rudiedirkx:
I agree with

> You can use custom template. I don't like the idea of defining these fields array.

Period. If it's only visual, it's a template. There's no other benefit in a new type (like validation or speed or readability) IMO.

(@yaquawa It's very confusing to combine PRs. For you it's easier for dev, but for reviewing it's confusing.)

yaquawa:
@rudiedirkx 
Thanks for commenting.

As I described above, you can't just simply use a custom template to achieve the same result, because if you do so, all the validation things won't be applied. 

I'd like to suggest using a method `group` to  warapping mutiple fields https://github.com/kristijanhusak/laravel-form-builder/pull/580#issuecomment-577670656
what do you think?

yaquawa:
> (@yaquawa It's very confusing to combine PRs. For you it's easier for dev, but for reviewing it's confusing.)

Yes, thanks for telling me this. I'm supposing #577 is going to be merged, and that PR changes the phpunit to modern version, so lots of code will be very different from this, it's very difficult to merge. I think I can rebase it on #577 later.

rudiedirkx:
Why won't validation work? Because of the child form nested name? Don't make it a child form, just make it separate fields in the same form. The grouping is only visual, isn't it? If the fields are reused elsewhere, we could make a mixin or something: `$this->mixin(MoreFieldsForm::class)`. Just for DRYness.

rudiedirkx:
I have a few `groups` like this, where I write out the entire row of separate fields (in Twig):

```
  <div class="form-item form-select form-inline auto-width duration">
    <label class="control-label required">{{ trans('educational_institution.variant.duration') }}</label>
    <div class="form-inline-inline-widgets">
      {{ form_widget(form.min_duration) }}
      <span>-</span>
      {{ form_widget(form.max_duration) }}
      <span></span>
      {{ form_widget(form.duration_type) }}
    </div>
    {{ form_errors(form.min_duration) }}
    {{ form_errors(form.max_duration) }}
    {{ form_errors(form.duration_type) }}
  </div>
```

That's 3 fields in a 'group'/row. The belong together, but not as a separate child form. They're all properties in the same form, with more fields (in separate rows).

No validation issues, just a visual grouping.

yaquawa:
@rudiedirkx 
Thanks for sharing your code.
But I think that's not a `template`, you are rendering the whole form manually instead of `form($form)`, right ? 
Maybe It's ok for small form, but think of if you have a very large form, the view may become messy very soon and it's unreusable. And if you use this method in collection and child form, the view become even more messy. 

@rudiedirkx @kristijanhusak 
The `group` method which I've suggested is not going to add the fields as child fields, but just regular field with a group tag so they can be rendered as group. If you guys are ok for this idea I can make another PR.

rudiedirkx:
Yes, that's a template. It's not a `form.php` template replacement, but the page template. I've created some helpers to make it very easy mostly. `form_row()`, `form_rows()`, `form_unless()`, `form_until()` etc. The full template is not much longer than the part I copied. The custom grouping is an exception. Most forms are way shorter, or just `form(form)`. I've added the full form code.

I still don't see how Form class grouping is better than template grouping... You can make a PR for the `group()` method, or just create a new package for the whole thing you need. The form builder should be pretty pluggable. If it's not pluggable enough, we'll fix it.

yaquawa:
@rudiedirkx I have a form that can be reused in multiple places with adding or removing some fields. It's much easier to just `form($form)` rather than including a blade template with many options, and in the blade template using `@if` to branch logic. And as I mentioned earlier, it become messy when you have a child form or a child form in collection. 

I'd like to make a PR to implement the `group` method, but you can reject it if you don't think this is a good idea anyway.

Issue #581, opened 2020-02-05
What's the purpose of the `Form::exclude` method ?


yaquawa:
I didn't see any benefits that can get from the `Form::exclude` method.

By calling this method, the form won't render the fields in the exclude list, but it will still check the rules of the fields in the exclude list, which don't make any sense for me. Why not just using the `Form::remove` method instead ?

Can anyone explain me the usage of the exclude method?

rudiedirkx:
I've never used it. I don't know it.. Don't think about it =) If it's not in the way, fuggetaboutit.

PR #582, opened 2020-02-05
Added a new method `Form::group`.

yaquawa:
Alternative PR to #580 which implemented a new `group` method.
It was branched from #577 you can compare the changes at [here](https://github.com/yaquawa/laravel-form-builder/compare/e0951c54f5ac7d581b9bb49155ee62dbfc7c8e6c...yaquawa:field_group)
(I'm going to rebase this on master once #577 get merged)

## Usage

```php
$form
    ->add('foo')
    ->group(function (FieldGroup $fieldGroup) {
        $fieldGroup->add('bar');
        $fieldGroup->add('foobar');
        $fieldGroup->group(function (FieldGroup $subFieldGroup) {
            $subFieldGroup->add('sub_bar');
            $subFieldGroup->add('sub_foobar');
        });
    })
    ->group(function (FieldGroup $fieldGroup) {
        $fieldGroup->add('baz');
    })
    ->add('qux');
```

Outputs:
```html
<form method="GET" action="http://localhost" accept-charset="UTF-8">
    <div class="form-group">
        <label for="foo" class="control-label">Foo</label>
        <input class="form-control" name="foo" type="text" id="foo">
    </div>
    <fieldset class="form-group">
        <div class="form-group">
            <label for="bar" class="control-label">Bar</label>
            <input class="form-control" name="bar" type="text" id="bar">
        </div>
        <div class="form-group">
            <label for="foobar" class="control-label">Foobar</label>
            <input class="form-control" name="foobar" type="text" id="foobar">
        </div>
        <fieldset class="form-group">
            <div class="form-group">
                <label for="sub_bar" class="control-label">Sub bar</label>
                <input class="form-control" name="sub_bar" type="text" id="sub_bar">
            </div>
            <div class="form-group">
                <label for="sub_foobar" class="control-label">Sub foobar</label>
                <input class="form-control" name="sub_foobar" type="text" id="sub_foobar">
            </div>
        </fieldset>
    </fieldset>
    <fieldset class="form-group">
        <div class="form-group">
            <label for="baz" class="control-label">Baz</label>
            <input class="form-control" name="baz" type="text" id="baz">
        </div>
    </fieldset>
    <div class="form-group">
        <label for="qux" class="control-label">Qux</label>
        <input class="form-control" name="qux" type="text" id="qux">
    </div>
</form>
```

### DIFF
--- a/src/Kris/LaravelFormBuilder/FieldGroup.php
+++ b/src/Kris/LaravelFormBuilder/FieldGroup.php
@@ -1,0 +1,228 @@
+<?php
+
+namespace Kris\LaravelFormBuilder;
+
+use Illuminate\Support\Arr;
+
+class FieldGroup
+{
+    /**
+     * The form that holds all the added fields.
+     *
+     * @var Form
+     */
+    protected $form;
+
+    /**
+     * The parent form.
+     *
+     * @var Form
+     */
+    protected $parent;
+
+    /**
+     * @var FormHelper
+     */
+    protected $formHelper;
+
+    /**
+     * All options for the group.
+     *
+     * @var array
+     */
+    protected $options = [];
+
+    /**
+     * The template name to be used.
+     *
+     * @var string
+     */
+    protected $template;
+
+    /**
+     * FieldGroup constructor.
+     * @param Form $parent
+     * @param array $options
+     */
+    public function __construct(Form $parent, array $options = [])
+    {
+        $this->parent = $parent;
+        $this->formHelper = $parent->getFormHelper();
+        $this->options = $this->formHelper->mergeOptions($this->getDefaults(), $options);
+        $this->form = (clone $parent)->removeAll();
+        $this->template = $parent->getConfig($this->getTemplate(), $this->getTemplate());
+    }
+
+    /**
+     * Get the template, can be config variable or view path.
+     *
+     * @return string
+     */
+    protected function getTemplate()
+    {
+        return 'group';
+    }
+
+    /**
+     * Default options for field.
+     *
+     * @return array
+     */
+    protected function getDefaults()
+    {
+        return [
+            'label' => null,
+            'label_attr' => ['class' => $this->parent->getConfig('defaults.label_class')],
+            'wrapper' => ['class' => $this->parent->getConfig('defaults.group.wrapper_class') ?? $this->parent->getConfig('defaults.wrapper_class')],
+        ];
+    }
+
+    /**
+     * @return Fields\FormField[]
+     */
+    public function getFields()
+    {
+        return array_intersect_key($this->parent->getFields(), $this->form->getFields());
+    }
+
+    public function hasField(string $fieldName)
+    {
+        return array_key_exists($fieldName, $this->getFields());
+    }
+
+    /**
+     * @param $name
+     * @param string $type
+     * @param array $options
+     * @return $this
+     */
+    public function add($name, $type = 'text', array $options = [])
+    {
+        $this->form->add($name, $type, ['fieldGroup' => $this] + $options);
+
+        return $this;
+    }
+
+    /**
+     * @param \Closure $callback
+     * @param FieldGroup|null $fieldGroup
+     * @return $this
+     */
+    public function group(\Closure $callback, FieldGroup $fieldGroup = null)
+    {
+        $this->form->group($callback, $fieldGroup, $this->parent);
+
+        return $this;
+    }
+
+    /**
+     * Render the field.
+     *
+     * @param array $options
+     * @return string
+     */
+    public function render(array $options = [])
+    {
+        $data = $this->getRenderData();
+
+        return $this->formHelper->getView()->make(
+            $this->getViewTemplate(),
+            $data + [
+                'group' => $this,
+                'options' => $this->formHelper->mergeOptions($this->options, $options),
+                'fields' => static::buildFieldsForRendering($this),
+            ]
+        )->render();
+    }
+
+    /**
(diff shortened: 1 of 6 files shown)


rudiedirkx:
You removed `if( ! in_array($field->getName(), $exclude) )`?

Why make a form copy, `compose()` its fields and discard it? Why not immediately add fields to the real form?

It feels a bit weird. It's like a `ParentType`, but done differently. `ParentType` should (be able to) do this. We don't need another kinda-`ParentType`. #580 seems perfect for a separate package. Why this?

I'm sorry. I feel like an asshole. I still don't get it, so I keep bitching. Templates, man =) They're so easy.

yaquawa:
@rudiedirkx Thanks for reviewing👍🏼
> You removed if( ! in_array($field->getName(), $exclude) )?

Yes, because I don't think this was even be used as I asked at #581...  I can bring it back if needed. @kristijanhusak What do you think?

> Why make a form copy, compose() its fields and discard it?

The cloned form object is for storing the added fields in the group. 

> Why not immediately add fields to the real form?

Yes, the added fields were [added to the real form too](https://github.com/kristijanhusak/laravel-form-builder/pull/582/commits/ef6029d2a79665b9b2540f5bbfabdd29c86b618e#diff-7f17d71f56dd1175c49875c83e9adc04R316).

> It feels a bit weird. It's like a ParentType, but done differently. ParentType should (be able to) do this. We don't need another kinda-`ParentType`. #580 seems perfect for a separate package. Why this?

I've done it by mimicking the `ParentType` on #580. At first I thought that's the way to go, but as @kristijanhusak said:

> Let's try to do something that will work in similar way as compose, but to allow setting up the container.

then I realized that making it looks like a parent can be very confusing (semantically)... keeping the hierarchy as it is is more intuitively(so we can `remove` or `modify` the fields in the group directly).

> I still don't get it, so I keep bitching. Templates, man =) They're so easy.

Hmm.. I like this package because I don't have to think about views.. I think that's the beauty of this package. And as I said before, if you are going to use a template instead, it comes quite messy very soon if you start using group in collection or child form field type. I'd like just `form($form)` and forget anything else. Again that's the reason I like this package. Grouping fields is very common, why not..?

kristijanhusak:
@yaquawa please clean up the PR so I can see only code relevant to this PR. I don't want to see rules PR here.

yaquawa:
@kristijanhusak You can see the diff [here](https://github.com/yaquawa/laravel-form-builder/compare/e0951c54f5ac7d581b9bb49155ee62dbfc7c8e6c...yaquawa:field_group), I used some phpunit methods that not exists in the legacy version of phpunit before #577.

kristijanhusak:
@yaquawa lets then have a PR that only updates phpunit so we can split all these PRs that you create to have only code related to the PR. I can't go through 20 files for each PR, i have no idea what was changed.

yaquawa:
Hi @kristijanhusak , I mean you can see the changes by comparing the #577 with this branch:

https://github.com/yaquawa/laravel-form-builder/compare/e0951c54f5ac7d581b9bb49155ee62dbfc7c8e6c...yaquawa:field_group

What you see is the same result even I created another PR to update phpunit.

Then after #577 got merged, this can be rebased on it.

rudiedirkx:
Yeah @kristijanhusak let's merge [#577](https://github.com/kristijanhusak/laravel-form-builder/pull/577#issuecomment-584140269) =p

yaquawa:
Thanks @rudiedirkx !
I've just rebased this PR onto the master👍🏼

rudiedirkx:
I don't know how to retry checks (Travis). Can Github do that?

kristijanhusak:
@rudiedirkx i retried it on travis.

rudiedirkx:
It seems to retry with the same config =) Which makes sense, because that's the config in this PR.

@yaquawa Can you do another rebase from `master`? That will take the right config and trigger tests.

yaquawa:
@rudiedirkx Done!

rudiedirkx:
Aw yeah look at that continuous integrationness

> All checks have passed
> 11 successful checks

`11` is just twice as much as `1`, right?

The PR is for @kristijanhusak to decide though. I'm against, but only slightly.

kristijanhusak:
@yaquawa this just looks like something that is used to avoid adding any custom rendering to view with some child forms. Intention of the package was never to make a form and just use `form($form)` all the time. It is possible sometimes, but for complicated forms, you should use other view helpers to achieve what you need. 
As the title of the package says, Symfony like form builder. Symfony form builder requires same amount, or even more logic in the views than this package.

If you really need this functionality, you can always add it as a custom field and just use it wherever you need.
I appreciate the effort to add this, but I think it's out of scope since it can be added as a customization.

yaquawa:
> this just looks like something that is used to avoid adding any custom rendering to view with some child forms.

You can specify a custom template for the group too.

SanderMuller:
I like this idea for something like bootstrap tabs, where you organize a form in groups and then in the template render each group in it's separate tab. 👍

mertasan:
This is exactly what I need. I'm having trouble making a template like the one below. Of course I can do it using the templates. However, if I were to do it manually, I wouldn't prefer to use a form builder. Please consider this issue.

```html
<div class="form-group row">
    <div class="col-lg-4">
        <input>
    </div>
    <div class="col-lg-4">
        <input>
    </div>
</div>
<div class="form-group row">
    <div class="col-lg-4">
        <input>
    </div>
    <div class="col-lg-4">
        <input>
    </div>
</div>
```

MahdiPishguy:
@mertasan could you resolve your problem?

PR #587, opened 2020-02-15
Add support for "allow destroy" to `ChildForm`

yaquawa:
Currently `collection` type with `'type' => 'form'` be set can only add child but can not destroy existing ones.

This PR adds an option `allow_destroy`(took the name from [Rails](https://api.rubyonrails.org/classes/ActiveRecord/NestedAttributes/ClassMethods.html)) to `ChildForm`.

With `allow_destroy` equals `true`, all fields got removed except the fields given by the `required_fields_for_allow_destroy` option. This can avoid unnecessary validations on the items which are going to be destroyed.

### DIFF
--- a/src/Kris/LaravelFormBuilder/Fields/ChildFormType.php
+++ b/src/Kris/LaravelFormBuilder/Fields/ChildFormType.php
@@ -38,7 +38,13 @@ class ChildFormType extends ParentType
             'value' => null,
             'formOptions' => [],
             'data' => [],
-            'exclude' => []
+            'exclude' => [],
+
+            // Add a field named `_destroy` to the child form if true.
+            'allow_destroy' => false,
+
+            // All the fields except the given fields will be removed if `allow_destroy` is `true`.
+            'required_fields_for_allow_destroy' => ['_destroy'],
         ];
     }
 
@@ -76,14 +82,20 @@ class ChildFormType extends ParentType
 
     /**
      * @return mixed|void
+     * @throws \Exception
      */
     protected function createChildren()
     {
-        $this->form = $this->getClassFromOptions();
+        $this->form = $this->getFormFromOptions();
+
+        if ($this->allowDestroy()) {
+            $this->enableAllowDestroy();
+        }
 
         if ($this->form->getFormOption('files')) {
             $this->parent->setFormOption('files', true);
         }
+
         $model = $this->getOption($this->valueProperty);
         if ($this->isValidValue($model)) {
             foreach ($this->form->getFields() as $name => $field) {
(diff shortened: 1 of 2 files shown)


rudiedirkx:
Eh. What? Why? I don't know what destroying a form or fields means... Is it like a delete button that posts the form? How would you use that?

yaquawa:
@rudiedirkx 
Here is an example it can be used on.

```php
<?php

class BankAccountForm extends Form
{
    public function buildForm()
    {
        $this
            ->add('ID')
            ->add('bank_name')
            ->add('account_number');
    }
}

class UserForm extends Form
{
    public function buildForm()
    {
        $this
            ->add('name')
            ->add('email')
            ->add('bank_accounts', 'collection', [
                'type' => 'form',
                'options' => [
                    'class' => BankAccount::class,
                    'label' => false,
                    
                    // Add this to allow items which has key `allow_destroy` equals to true to be ignored from validation.
                    'allow_destroy' => true, 
                    
                    // Only these fields will be kept.
                    // Which means `bank_name and `account_number` will be removed from the child from.
                    // '_destroy' and `ID` will be kept
                    // We need '_destroy' so we know which item should be removed.
                    // We need `ID`(the database key name of the item) too to identify the item from DB.  
                    'required_fields_for_allow_destroy' => ['ID', '_destroy']
                ]
            ]);
    }
}
```

rudiedirkx:
So what's the goal? A delete button?

What do you do with it? Does the form/template html work automatically? What does the controller do with it?

yaquawa:
> What do you do with it?

To provide a "delete item" feature. I'm wondering that haven't any of you needed this feature so far? If yes, how will you implement that? I thought this is the proper way to go.

> Does the form/template html work automatically? 

That should be the js domain's responsibility, isn't it?

> What does the controller do with it?

The controller's responsibility would be deleting the items which has the `_destroy` set to `true`.

rudiedirkx:
We haven't needed it because it's a weird way to delete a collection item IMO. I would just delete the item with JS, and save the form. Missing items are deleted, other items are updated/created, the rest of the form is saved. Multiple submit buttons in multiple locations seems weird to me. Maybe I'm weird.

kristijanhusak:
@yaquawa i'll have to agree with @rudiedirkx on this one. I assume you need something like this to be able to distinguish which collection entry needs to be destroyed, and which not. You can easily add a hidden field yourself and manage it however you want, there's no need to add this functionality as part of package.

yaquawa:
@kristijanhusak 

> You can easily add a hidden field yourself and manage it however you want

But how could I remove all other fields if the destroy field equals true ?
Can you tell me how have you guys implemented the "destroy item" feature before ?

rudiedirkx:
I know I'm crazy late, but this might still be relevant.

> Can you tell me how have you guys implemented the "destroy item" feature before ?

I give my collection forms an `id`/`ID` field which is automatically filled by the form builder. I set `prefer_input = true`. Then I let JS add and delete collection items. Then in POST I know which `id`'s exist in the post data, so I know not to delete those. All `id` that don't exist in the post data anymore (but do exist in the model's relationship) have been deleted in JS, so I know to delete those relationships.

In the POST action:

```
$values = $form->getFieldValues();
$currentIds = $user->bank_accounts->pluck('id')->all();
$keepIds = array_column($values['bank_accounts'], 'id');
$deleteIds = array_diff($currentIds, $keepIds);
$user->bank_accounts->whereIn('id', $deleteIds)->each->delete();
```

PR #588, opened 2020-02-15
Field name as default prototype name for `CollectionType`

yaquawa:
Make the field name as default prototype name for `CollectionType`, so we don't need set the prototype name manually whenever we add a collection field.

### DIFF
--- a/src/Kris/LaravelFormBuilder/Fields/CollectionType.php
+++ b/src/Kris/LaravelFormBuilder/Fields/CollectionType.php
@@ -2,6 +2,7 @@
 
 namespace Kris\LaravelFormBuilder\Fields;
 
+use Illuminate\Support\Str;
 use Illuminate\Support\Collection;
 
 class CollectionType extends ParentType
@@ -31,13 +32,15 @@ class CollectionType extends ParentType
      */
     protected function getDefaults()
     {
+        $prototypeName = Str::camel($this->getRealName());
+
         return [
             'type' => null,
             'options' => ['is_child' => true],
             'prototype' => true,
             'data' => null,
             'property' => 'id',
-            'prototype_name' => '__NAME__',
+            'prototype_name' => "__{$prototypeName}__",
             'empty_row' => true,
             'prefer_input' => false,
         ];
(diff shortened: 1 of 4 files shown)


rudiedirkx:
Do you mean "set the prototype name manually" in JS, when adding the dynamic field? What happens if you add 3 fields? They can't all get the same name. It has to be manually.

yaquawa:
> It has to be manually.
No it can be "automatically" by setting the prototype name to a random value, but I'd like to set it to the field name(which this PR does so).

rudiedirkx:
What happens if you add 3 fields with JS? You use the same name 3 times. That doesn't work. That's why it has to be JS that sets the name, and that's why there's a `__NAME__` placeholder in the prototype.

yaquawa:
@rudiedirkx  
> add 3 fields with JS

Did you mean :
1. you click a "add item" button
2. new item added
3. repeat above steps 3 times.

rudiedirkx:
Yes, that's what I mean. Without a `__NAME__` placeholder, JS will add the same names 3 times, right? So JS has to do the replacement, not PHP.

kristijanhusak:
@yaquawa this is ok for me, but we need to figure out how to avoid breaking backward compatibility. If we release this, Javascript that people use to replace the `__NAME__` with the index number will break. It can maybe be a configuration that defaults to false, something like `'use_form_name_as_prototype' => false`, and you can override it for your project if you want.

yaquawa:
> So JS has to do the replacement, not PHP.

But that's how the current version works isn't it ? Is there any deference?

yaquawa:
> @yaquawa this is ok for me, but we need to figure out how to avoid breaking backward compatibility. If we release this, Javascript that people use to replace the `__NAME__` with the index number will break. It can maybe be a configuration that defaults to false, something like `'use_form_name_as_prototype' => false`, and you can override it for your project if you want.

How about add a `Form::version` property? If versions before this PR, use `__NAME__` as default.

rudiedirkx:
Why don't you just manually set the `prototype_name` option to whatever your field name is? It's always a few lines above, right? Never in a separate file or function?

rudiedirkx:
@yaquawa ?

PR #591, opened 2020-03-04
Add support Laravel 7

sangnguyenplus:
Hi @kristijanhusak,
Laravel 7 is released and I've tested your package on my app using Laravel 7. It's still working fine.
There is no change related to your package in Laravel 7 so I think you should support this version.
Thanks for the great package!


kristijanhusak:
Thanks!

PR #596, opened 2020-04-26
Allow use of the property option to resolve value

SanderMuller:
**What has changed?**
Allows the model attribute name to be defined as option to resolve the value from the model, useful for when an accessor is used to modify the value on retrieval, but it should be stored in the database as the field name.

**Personal use case example**
```
//old method
$form->add('coverages', 'choice', [
    'selected' => $this->getModel()->form_coverages,
]);

//new method
$form->add('coverages', 'choice', [
    'property' => 'form_coverages',
]);
```

What did this fix for me?
Being able to use `setValue` on the form field afterwards. With hard defining the selected (default) value, the `setValue` method becomes unusable since it early returns if the field has a default value.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,9 +19,9 @@
     ],
     "require": {
         "php": ">=7.1",
-        "laravelcollective/html": "5.6.*|^6|^7",
-        "illuminate/database": "5.6.*@dev|^6|^7",
-        "illuminate/validation": "5.6.*@dev|^6|^7"
+        "laravelcollective/html": "^5.6|^6|^7",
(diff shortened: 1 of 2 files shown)


rudiedirkx:
Shouldn't include code from #597. 

I like it, but are you sure this is the only place where field's `name` should be replaced by option `property`?

rudiedirkx:
And can you add a simple test to read from an explicit `property`?

SanderMuller:
Closed in favor of https://github.com/kristijanhusak/laravel-form-builder/pull/598

PR #597, opened 2020-04-26
Fixes issue where Laravel 5.7 and 5.8 were no longer allowed

SanderMuller:

PR #598, opened 2020-04-26
Allow use of the property option to resolve value

SanderMuller:
**What has changed?**
Allows the model attribute name to be defined as option to resolve the value from the model, useful for when an accessor is used to modify the value on retrieval, but it should be stored in the database as the field name.

**Personal use case example**
```
//old method
$form->add('coverages', 'choice', [
    'selected' => $this->getModel()->form_coverages,
]);

//new method
$form->add('coverages', 'choice', [
    'property' => 'form_coverages',
]);
```

What did this fix for me?
Being able to use `setValue` on the form field afterwards. With hard defining the selected (default) value, the `setValue` method becomes unusable since it early returns if the field has a default value.

This is an update/cleanup from https://github.com/kristijanhusak/laravel-form-builder/pull/596 to include a test case and exclude the composer.json change from https://github.com/kristijanhusak/laravel-form-builder/pull/597

### DIFF
--- a/src/Kris/LaravelFormBuilder/Fields/FormField.php
+++ b/src/Kris/LaravelFormBuilder/Fields/FormField.php
@@ -142,7 +142,8 @@ abstract class FormField
         }
 
         if (($value === null || $value instanceof \Closure) && !$isChild) {
-            $this->setValue($this->getModelValueAttribute($this->parent->getModel(), $this->name));
+            $attributeName = $this->getOption('property', $this->name);
+            $this->setValue($this->getModelValueAttribute($this->parent->getModel(), $attributeName));
         } elseif (!$isChild) {
             $this->hasDefault = true;
         }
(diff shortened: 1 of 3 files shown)


rudiedirkx:
You changed existing tests to test a new feature... Does `$form->setModel()` not work anymore? That's why it's tested. That should still work. With this feature too.

SanderMuller:
The `setModel` refactor in those tests was not a functional change, it has no effect / relation to the new feature, and the tests pass with or without that change. I updated those two tests because they used the deprecated method of setting the model.

Perhaps there should be a separate tests to verify the deprecated method `setModel` itself works. There are still plenty of tests that use this deprecated method, for example `FormTest@it_can_set_form_options_with_setters`

I'm not a big fan of using deprecated code outside of special tests to verify that the deprecated code still works.

rudiedirkx:
Fair enough. I didn't even know `setModel()` was deprecated. I don't use a fancy enough IDE apparently.

PR #599, opened 2020-05-02
Fixes issue where the entity field was not set to the model value

SanderMuller:
We should not use the `property` option on an instance of the `EntityType` field, because it uses the `property` option to resolve the display name of the field. 

There was no unit test that validated that the value of the entity in the existing model was set as the value of the entity field.

This PR fixes the issue for the `EntityType` field and adds the unit test to validate this works.


rudiedirkx:
I never use 'entity' fields, but I believe you, and it can't break something existing.

PR #607, opened 2020-07-20
Added helperBlockPath() instead of hardcoded helper_block path

azisyus:
This PR may help to who wants to use helper_block.php and errors.php while creating custom inputs.

### DIFF
--- a/src/helpers.php
+++ b/src/helpers.php
@@ -4,6 +4,16 @@ use Kris\LaravelFormBuilder\Fields\FormField;
 use Kris\LaravelFormBuilder\Form;
 
 
+if(!function_exists('errorBlockPath'))
+{
+
+    function errorBlockPath()
+    {
+        return __DIR__.'/views/errors.php';
+    }
+
+}
+
 if(!function_exists('helpBlockPath'))
 {
 
--- a/src/helpers.php
+++ b/src/helpers.php
@@ -3,6 +3,17 @@
 use Kris\LaravelFormBuilder\Fields\FormField;
 use Kris\LaravelFormBuilder\Form;
 
+
+if(!function_exists('helpBlockPath'))
+{
+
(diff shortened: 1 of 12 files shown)


azisyus:
Let's say i have a X project which depends to Y package and Y is also depends to this package and my custom input is actually used in Y, the problem is helper_block cannot be used without moving my custom view into `base_path('resources/views/vendor/laravel-form-builder')` which may cause to manual intervention for different projects because my custom view will be in another package's published directory and it should be.

kristijanhusak:
Makes sense. Thanks.

rudiedirkx:
Even better would be if it were intelligent enough to find the most relevant view file. If an original view is executed, it might include an app overridden help_block. Or the oppositve: an app overridden view includes the original help_block. Before this PR, both files (e.g. `select.php` and `help_block.php` had to be in the same dir. After this PR, the copied `select.php` doesn't even try to load the copied/custom `help_block.php`. This is not better.

rudiedirkx:
`errorBlockPath()` and `helpBlockPath()` should check if the view exists in `ROOT/resources/views/vendor/laravel-form-builder` or wherever the copied view `.php` files are copied to. That fixes both cases: original `select` with custom `help_block`, and custom `select` with original `help_block`.

azisyus:
Thats right, those functions doesn't load most relevant helper_block file which is bad but i believe we need better than just one or two hacky functions. Wouldn't be better if we can load `helper_block` with namespaced `@include`? We can pass `$options` as parameter to that `@include` directive, that would be better fix i suppose.

rudiedirkx:
I don't know what "namespaced `@include`" means. Is that a PHP thing? PHP's `include` doesn't take arguments as far as I know, and the include can't be in a function, because it needs the 'global'-ish template vars. All addressed in #608 ?

PR #608, opened 2020-07-22
Load error and help_block from package namespace instead of hardcoded path

azisyus:
This PR fixes a bug caused by my last PR (#607), view paths generated from package's namespace so we can use most relevant view.

### DIFF
--- a/src/helpers.php
+++ b/src/helpers.php
@@ -3,13 +3,26 @@
 use Kris\LaravelFormBuilder\Fields\FormField;
 use Kris\LaravelFormBuilder\Form;
 
+if(!function_exists('getFormBuilderViewPath'))
+{
+    function getFormBuilderViewPath($fileName)
+    {
+        $p = explode('.',$fileName);
+        $c = count($p);
+        if($c>2 || $p[$c-1]!='php')
+            throw new Exception('You should use only *.php files with this function');
+
+        $path = base_path('resources/views/vendor/laravel-form-builder/'.$fileName);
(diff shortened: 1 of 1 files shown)


rudiedirkx:
Does that work if I convert my views to Twig or Blade? I think `view($path)` will find any extension view, not just `.php`, and you can't just include a Twig or Blade file into PHP.

rudiedirkx:
I think the functions can be dumber. Just check if the `.php` file exists in the configured folder. I don't know where the folder is configed, but I'm sure `app()` or `view()` or `config('view')` knows.

azisyus:
`view($path)` doesn't load files with `*.twig` extension so you can't use that template engine but `*.blade.php` is fine because `view($path)` probably checks for that and it's only valid for `.blade.php` and `.php` files. Does this package promises usage of another template engine?

rudiedirkx:
> view($path) doesn't load files with *.twig extension so you can't use that template engine

My projects use Twig, so it does. That's how Laravel works. Any engine works. Laravel can render all those supported types, but php can't just include all those types.

Since Laravel supports any tpl engine, we must at least make sure it doesn't break. We don't have to look for `help_block.twig` and render it correctly IMO, but we must not include it like it's a `.php` file.

I would hard-check to only accept `.php`, not even `.blade.php`. Those files come from this package, so we support that. All others meh.

azisyus:
>  Laravel can render all those supported types, but php can't just include all those types.

That makes sense, thanks for you patience.

rudiedirkx:
I like it!

PR #612, opened 2020-08-19
Update CollectionType.php

bhcosta90:
Fix to correct one to many collect

### DIFF
--- a/src/Kris/LaravelFormBuilder/Fields/CollectionType.php
+++ b/src/Kris/LaravelFormBuilder/Fields/CollectionType.php
@@ -122,7 +122,16 @@ class CollectionType extends ParentType
         if (!$data || empty($data)) {
             if ($empty = $this->getOption('empty_row')) {
                 $val = $empty === true ? null : $empty;
-                return $this->children[] = $this->setupChild(clone $field, '[0]', $val);
+                if (count($currentInput) > 0) {
+                    foreach (array_keys($currentInput) as $k) {
+                        if (is_numeric($k)) {
+                            $this->children[] = $this->setupChild(clone $field, '['.$k.']', $val);
+                        }
+                    }
+                    return $this->children;
+                } else {
+                    return $this->children[] = $this->setupChild(clone $field, '[0]', $val);
+                }
             }
 
             return $this->children = [];


rudiedirkx:
Why? How? When? Etc. I'm very dumb. Explain to me.

bhcosta90:
This change solved my one to many problem when I was using Form Collect, I will put the example below

```
<?php

namespace Modules\Orcamento\Forms;

use Modules\Cliente\Entities\Cliente;

class OrcamentoForm extends Form
{
    public function buildForm()
    {
        $clientes = Cliente::all()->pluck('nome', 'id')->toArray();

        $this->add('cliente_id', 'select', [
            'label' => "Cliente",
            'choices' => ["" => "Selecione..."] + $clientes,
            'rules' => [
                'required'
            ]
        ])->add('contato_id', 'select', [
            'label' => "Contato",
            'choices' => ["" => "Selecione..."],
            'attr' => [
                'disabled' => true,
            ],
        ])->add('itens', 'collection', [
            'type' => 'form',
            'property' => 'name',
            'prototype' => true,            // Should prototype be generated. Default: true
            'prototype_name' => '__NAME__', // Value used for replacing when generating new elements from prototype, default: __NAME__
            'options' => [
                'label' => false,
                'attr' => ['class' => 'tag'],
                'class' => OrcamentoItemForm::class,
            ]
        ]);
    }
}

```

```
<?php

namespace Modules\Orcamento\Forms;

use Modules\Cliente\Entities\Cliente;

class OrcamentoItemForm extends Form
{
    public function buildForm()
    {
        $this->add('nome', 'text', [
            'label' => "Nome",
        ])->add('model_id', 'hidden', [
            'label' => "Model ID",
        ])->add('model_type', 'hidden', [
            'label' => "Model Type",
        ])->add('quantidade', 'number', [
            'label' => "Quantidade",
        ])->add('preco', 'number', [
            'label' => "Preço",
        ]);
    }
}

```

Before he only saved the first item, with my change he will save all the items in the array.

rudiedirkx:
Looks like a config error, not a package bug. The whole point of the Collection is to save more items, so it'd be very strange if it never did. Everyone would be complaining about this.

I'll explain, because the docs are still not good enough (and I'm still too lazy):

- ALWAYS build the form in the POST action (save) like you built it in the GET action (show form). That means the same model, the same data, the same parameters. Everything exactly the same. That way, the form validator knows how many items to validate & accept.
- If your POST data can contain more items than the GET form (e.g. because you add items with JS), set the `collection` config `prefer_input => true`.

I think the first applies to you, not the second. If that doesn't work, show me the controller actions where the forms are initiated.

rudiedirkx:
Also, why do you have a `'property' => 'name'`? Don't you want the whole child form?

rudiedirkx:
I've FINALLY created a chapter for this in the docs. Let me know if it's not perfect: https://kristijanhusak.github.io/laravel-form-builder/field/collection.html#validated-collection-items

Issue #614, opened 2020-08-25
Regex Rules aren't supported anymore


jelaucv:
ErrorException: **preg_match(): No ending delimiter '/' found** - the exception is thrown when a regex with '|' is used, even if rules are declared separately in array

jelaucv:
https://github.com/kristijanhusak/laravel-form-builder/pull/615

PR #615, opened 2020-08-25
Fix Regex/Array Rules normalisation

jelaucv:
Use same behaviour for rules as in Laravel Validator

Issue: https://github.com/kristijanhusak/laravel-form-builder/issues/614


kristijanhusak:
Will this break older laravel versions?

jelaucv:
> Will this break older laravel versions?

that shouldn’t break older laravel versions, laravel 5,6 supports rules declaration as pipeline string or only separately in an array

kristijanhusak:
Ok, thanks.

Issue #620, opened 2020-09-23
count(): Parameter must be an array or an object that implements Countable - Laravel 8.0


superlogicabrunocosta:
count(): Parameter must be an array or an object that implements Countable

![image](https://user-images.githubusercontent.com/59574132/94066270-a01d9f00-fdc2-11ea-8bb2-864097935f9b.png)

rudiedirkx:
We need much more context than a screenshot if you want us to help. I'm afraid you'll have to debug it. Where does this non-countable value come from? How do you create the form in the controller? With what data? What's the POST data? Etc.

superlogicabrunocosta:
Então, eu tenho dois formularios, um deles chama o outro via coleção, porem quando eu tenho dados o mesmo da esse erro. Criei um repositorio com esse simples teste para dar uma olhada

https://github.com/bhcosta90/laravel-8.git

superlogicabrunocosta:
So, I have two forms, one of them calls the other via collection, however when I have data the same as this error. I created a repository with this simple test to take a look

https://github.com/bhcosta90/laravel-8.git

rudiedirkx:
Ooh, an entire repo for reproduction. Interesting.

----

It's a bug in `FormField::setupValue()`. `$attributeName` is `id` instead of `emails`, so it takes the data/model's `id` as value (`1`, not countable), instead of the `emails` property, which is a list.

My fault in #598 and later with [typical famous last words](https://github.com/kristijanhusak/laravel-form-builder/pull/599#issuecomment-633106973). I'm ashamed. I should have tested that. And/or we're missing tests...

A workaround for now is very simple. Add `'property' => 'emails',` to the collection field settings (first level, not in `'options'`).

Real fix later. Or you can take a shot at it, but try not to break #598.

rudiedirkx:
Bug details and fix and discussion in #621

PR #621, opened 2020-09-27
Fixes bug from #620 introduced in #598: general field option 'property' clashes with collection field

rudiedirkx:
Bug introduced in #598 because the `property` option already exists in `collection` fields with default value `'id'`, so it **always** takes the `id` property for collection fields, which is always wrong.

Bug only noted in #620, because I suck, and automated tests didn't catch above bug.

Only fix IMO: rename the general `property` option to `value_property`, so it doesn't clash with the `collection` field option. Can't rename that one, because it's much older and officially released. Sorry @SanderMuller but that's a breaking change for you.

Added a CollectionType test to make sure the correct value is used.

I need a professional opinion here @kristijanhusak because I screwed this up before.

I will update the docs for `value_property` after merge.

### DIFF
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -7,7 +7,7 @@
          convertWarningsToExceptions="true"
          processIsolation="false"
          stopOnFailure="false"
-         syntaxCheck="false">
+    >
     <testsuites>
         <testsuite name="Package Test Suite">
             <directory suffix=".php">./tests/</directory>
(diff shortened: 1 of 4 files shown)


kristijanhusak (approved):

rudiedirkx:
Thanks @kristijanhusak.

@SanderMuller I don't even know if you get these notifications, but if you do: read this issue.

SanderMuller:
@rudiedirkx thank you for tagging me. I was using a custom fork for earlier Laravel 7 support, but the tag helped me move back to this repo and resolve the breaking change. My apologies for introducing the bug in the first place.

Issue #636, opened 2021-02-26
Choice doesn't render correct


volovodenko:
```
class ChangeEvaluationPeriodStageForm extends Form
{
    protected $formOptions = [
        'id' => 'change_stage',
    ];

    public function buildForm()
    {
        $this->add('stage', 'choice', [
            'rules'          => ['required', 'string', Rule::in(EvaluationPeriodStage::values())],
            'choices'        => EvaluationPeriodStage::getSelectArray(),
            'expanded'       => true,
            'multiple'       => false,
            'choice_options' => [
                'wrapper'    => ['class' => null],
                'label_attr' => ['class' => null],
                'attr'       => ['class' => null],
            ],
            'label' => false,
        ]);

        $this->add('submit', 'submit', [
            'label' => trans('EvaluationPeriod::evaluation_periods.change_stage.submit.label'),
            'attr'  => ['class' => 'btn btn-block btn-primary mt-4'],
        ]);
    }
}
```

![image](https://user-images.githubusercontent.com/29176940/109328664-5f8aeb80-7862-11eb-9786-1395067c4697.png)

![image](https://user-images.githubusercontent.com/29176940/109329014-c6100980-7862-11eb-8f6a-5540c7d2530c.png)

rudiedirkx:
Yeah, that's a bug. Wanna crush it? I don't use `choice`, so I'm not very motivated.

PR #637, opened 2021-02-28
Fix choice render, #636

volovodenko:
Fix choice render, #636

### DIFF
--- a/src/Kris/LaravelFormBuilder/Fields/FormField.php
+++ b/src/Kris/LaravelFormBuilder/Fields/FormField.php
@@ -311,11 +311,16 @@ abstract class FormField
             if ($this->parent->clientValidationEnabled()) {
                 $this->setOption('attr.required', 'required');
             }
+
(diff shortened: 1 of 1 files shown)


rudiedirkx:
Why the second part? The foreach + set single options instead of setOption array once. Because `$parsedRules` is prefered over `$this->getOption('attr')`? Then you could switch the args in `$this->getOption('attr') + $parsedRules` for the same? Not that that's better, I just want to understand your change.

volovodenko:
For ChoiceType class, method setOption in FormField class is referenced to setOption in ParentType class, and then in if condition we will replace all 'attr' for children options, instead of merging with them. So we need to pass exact attribute name in FormField setOption method, or maybe fix setOption in ParentType class

rudiedirkx:
Right. That's why I don't use ChoiceType. Every option is copied to all children. Silly.

But this works! Thanks a lot!

PR #638, opened 2021-03-13
Use collection empty_model option for keeping model data intact during validation (errors)

rudiedirkx:
This is the better version of #519, with the same intent:

- Keep model input as models during validation errors, so child forms always have a predictable model object as `$this->model`, instead of an array.
- Make all not-model collection rows from a predictable real model (like prototype and empty row), which currently get the parent model as the child model, which is just weirdd.

Usage:

```
class SiteCostsForm extends Form {
	public function buildForm() {
		$this->add('costs', 'collection', [
			'type' => 'form',
			'prototype' => true,
			'prefer_input' => true,

			'empty_model' => fn() => new SiteCost(), // <<<

			'options' => [
				'class' => SiteCostForm::class,
			],
		]);
	}
}
```

`empty_model` will be run for every row that doesn't have a model:

- empty_row
- prototype
- **new** rows from form input (`$oldInput`) during validation

Existing rows during validation will get their original model object. That's new too. `formatInputIntoModels()` combines form input and db data.

@kristijanhusak I'm pretty sure about this one, except for one thing: `generatePrototype()`. It used to distinguish between `null` and `false`, but I can't find any different anywhere. Both `null` and `false` would use the parent model in the child form. With this change it uses `empty_model` or null (the parent model).

### DIFF
--- a/src/Kris/LaravelFormBuilder/Fields/CollectionType.php
+++ b/src/Kris/LaravelFormBuilder/Fields/CollectionType.php
@@ -104,14 +104,14 @@ class CollectionType extends ParentType
             $data = $currentInput;
         }
 
+        if ($data instanceof Collection) {
+            $data = $data->all();
+        }
+
         // Needs to have more than 1 item because 1 is rendered by default.
         // This overrides current request in situations when validation fails.
         if ($oldInput && count($oldInput) > 1) {
             $data = $this->formatInputIntoModels($oldInput, $data);
-        }
-
-        if ($data instanceof Collection) {
-            $data = $data->all();
         }
 
         $field = new $fieldType($this->name, $type, $this->parent, $this->getOption('options'));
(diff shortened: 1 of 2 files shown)


kristijanhusak:
@rudiedirkx Looks ok. Could you please write a test for this if we don't have one already?

rudiedirkx:
I was afraid you might say that. I don't know how to test form internals during validation error.... I can't just test the form output, because this is about what's inside the form (the child form's `$this->model` mostly).

rudiedirkx:
Hehheh found the first bug. Second commit incoming! Maybe even with tests, because apparently we can fake 'old input'.

kristijanhusak (approved):

kristijanhusak:
Looks good. I'll let you merge it, in case you want to add something else.

rudiedirkx:
Excellent. Next: update test suite to actually use modern versions of PHP and Laravel.

PR #644, opened 2021-05-07
choice checkboxes required attributes bugfix

rudiedirkx:
Never copy the `required` attr to a ChoiceType's children if they're checkboxes. For select or radiobuttons it's okay, but not for checkboxes, because that makes every individual checkbox required.


rudiedirkx:
This is horrible. There has to be a better way.

PR #647, opened 2021-07-02
array_filter function removed

abhitheawesomecoder:
array_filter function removed because, If no callback is supplied, all empty entries of array will be removed.


rudiedirkx:
I think that's the point... Only fields with filters are collected to filter input. What would be the point of including fields that don't have filters?

I don't use field filters, but I think the `array_filter` makes sense.

PR #653, opened 2021-11-18
Change mergeOptions() to merge instead of replacing by key.

brenno263:
In my use case, we've extended FormBuilder\Fields\SelectType and we're supplying values for $options['rules'] and $options['rules_append'] in in getDefaults( ) and the consructor respectively, but these are being removed if there are values set in $form->add(... 'rules' => [...]).

This is because on FormField line 273, we use this helper to merge options from the class and those from $form->add(...). array_replace_recursive(), however, replaces based on key value. And key value is just basic integers. Instead, we need to array_merge_recursive().

### DIFF
--- a/src/Kris/LaravelFormBuilder/FormHelper.php
+++ b/src/Kris/LaravelFormBuilder/FormHelper.php
@@ -137,7 +137,7 @@ class FormHelper
      */
     public function mergeOptions(array $targetOptions, array $sourceOptions)
     {
-        return array_replace_recursive($targetOptions, $sourceOptions);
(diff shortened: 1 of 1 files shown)


rudiedirkx:
I kind of agree, BUT that's a huge BC break, because `array_merge_recursive` and `array_replace_recursive` are significantly different. I think your use case is stranger than the normal use case, so this wouldn't be a good change. Maybe we can make your use case easier, but not with a huge BC break. Can you add examples?

rudiedirkx:
When #658 gets in, you can easily extend the form helper and override `mergeOptions()`.

PR #656, opened 2021-12-20
Getting field type by class name

ukeloop:
Getting field type by class name that is like Symfony.

```php
$form->add('name', EmailType::class);
```

```php
class EmailType extends InputType
{
    public function getType()
    {
        return 'email';
    }
}
```

### DIFF
--- a/src/Kris/LaravelFormBuilder/Fields/FormField.php
+++ b/src/Kris/LaravelFormBuilder/Fields/FormField.php
@@ -517,6 +517,18 @@ abstract class FormField
     }
 
     /**
+     * Marks the view as rendered.
+     *
+     * @return $this
+     */
+    public function setRendered()
+    {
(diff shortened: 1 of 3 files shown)


rudiedirkx:
I like the feature/fix, BUT it seems like you're prefering `availableFieldTypes` over class name. So you could define a field type `App\Fields\MyField` to load class `App\Forms\Fields\AnotherField`, even if they both exist. Maybe that's intentional, to override a specific class reference to another class, but that seems future-buggy. Maybe we should always prefer the class name (if exists), whatever `availableFieldTypes` says. @kristijanhusak ?

ukeloop:
In my opinion, there is no batching because the class name contains namespace.

ukeloop:
Added check subclass and fixed bugs.

PR #657, opened 2021-12-20
Added `setRendered` method to `FormField` class.

ukeloop:
We need the function of marking the view as rendered that is like Symfony.

https://github.com/symfony/symfony/blob/e7142a6651bedb33dcce7fb33d2895e3032764e4/src/Symfony/Component/Form/FormView.php#L83

### DIFF
--- a/src/Kris/LaravelFormBuilder/Fields/FormField.php
+++ b/src/Kris/LaravelFormBuilder/Fields/FormField.php
@@ -517,6 +517,18 @@ abstract class FormField
     }
 
     /**
+     * Marks the view as rendered.
+     *
+     * @return $this
+     */
+    public function setRendered()
+    {
+        $this->rendered = true;
+
+        return $this;
+    }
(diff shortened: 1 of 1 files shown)


rudiedirkx:
Why? How? Why?

ukeloop:
It's for modding without template in each view.

e.q. Mode parent field in blade

```blade
@php $form->foo->setRendered() @endphp

@foreach($form->foo->getChildren() as $child)
    {{ $child->getOption('value') }}
    {{ $child->getOption('label') }}
@endforeach
```

rudiedirkx:
That seems like an odd hack. If you're doing the `foreach` yourself, you could exclude `foo` yourself too. Why like this? To not append underendered field at the end? Or automatically in `ParentType` fields? I need more context.

Did you check all `isRendered()` calls? I can't estimate how this change would benefit OR risk me...

PR #658, opened 2021-12-20
Customizable classes by config.

ukeloop:
Customizable classes by config.

e.g.

```php
'form_builder' => \Kris\LaravelFormBuilder\FormBuilder::class,
'helper' => \App\Form\MyFormHelper::class,
```

```php
class MyFormHelper extend FormHelper
{
    public function mergeOptions(array $targetOptions, array $sourceOptions)
    {
        return $targetOptions + $sourceOptions;
    }
}
```

### DIFF
--- a/src/Kris/LaravelFormBuilder/FormBuilderServiceProvider.php
+++ b/src/Kris/LaravelFormBuilder/FormBuilderServiceProvider.php
@@ -35,10 +35,10 @@ class FormBuilderServiceProvider extends ServiceProvider
 
         $this->app->singleton('laravel-form-builder', function ($app) {
 
-            return new FormBuilder($app, $app['laravel-form-helper'], $app['events']);
+            return new config('laravel-form-builder.form_builder', FormBuilder::class)($app, $app['laravel-form-helper'], $app['events']);
         });
 
-        $this->app->alias('laravel-form-builder', 'Kris\LaravelFormBuilder\FormBuilder');
+        $this->app->alias('laravel-form-builder', config('laravel-form-builder.form_builder', FormBuilder::class));
 
         $this->app->afterResolving(Form::class, function ($object, $app) {
             $request = $app->make('request');
@@ -62,10 +62,10 @@ class FormBuilderServiceProvider extends ServiceProvider
 
             $configuration = $app['config']->get('laravel-form-builder');
 
-            return new FormHelper($app['view'], $app['translator'], $configuration);
+            return new config('laravel-form-builder.form_builder', FormHelper::class)($app['view'], $app['translator'], $configuration);
         });
 
-        $this->app->alias('laravel-form-helper', 'Kris\LaravelFormBuilder\FormHelper');
+        $this->app->alias('laravel-form-helper', config('laravel-form-builder.form_builder', FormHelper::class));
     }
 
     /**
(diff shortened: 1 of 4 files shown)


rudiedirkx:
- Definitely don't use `config()` in a service provider.
- Don't **remove** the standard alias `Kris\LaravelFormBuilder\FormBuilder`. You can add the custom form builder class as alias, but don't remove the package form builder class alias,

ukeloop:
What are the reasons for them? Many libraries have provided hacks for them.

If I mode helper method, I need override the service provider.

rudiedirkx:
Reasons for what? Hacks for what?

You shouldn't use `config()` in a SP because you have `$this->app['config']` right there.

You shouldn't remove the standard alias, only add others, because others might use that standard alias, so it still has to work. So the `$this->app->alias()` bit should check if the configed class is the standard class, and then add 1 or 2 aliases.

rudiedirkx:
That's not what I meant. My worry is that there's no alias for `Kris\LaravelFormBuilder\FormBuilder` if you change the form builder class to `App\MyFormBuilder`. `laravel-form-builder` and `App\MyFormBuilder` will work, but `Kris\LaravelFormBuilder\FormBuilder` won't work anymore (**I think**) because it's not an alias anymore. Any package will expect `Kris\LaravelFormBuilder\FormBuilder` to work, because it doesn't know if the dev changed the form builder class (or even that that's possible now).

I'm gonna do some tests with this branch.

rudiedirkx:
Yes, the standard alias is 'broken'. You can't 'constructor inject' with the container's `Kris\LaravelFormBuilder\FormBuilder` if you've overridden the form builder class.

And you should add `_class` to the config names, so we can also do this for the form class (for plain forms). `form` is already taken for the template.

rudiedirkx:
So it should

    $this->app->alias($abstract, $formBuilder);
    $this->app->alias($abstract, FormBuilder::class);

and then it doesn't matter how you ask the DIC for the form builder: 'laravel-form-builder', `Kris\LaravelFormBuilder\FormBuilder` and `App\MyFormBuilder` will all get the DIC's `App\MyFormBuilder` object.

And the same for the helper.

And I would love it if you did the form class too =) FormBuilder already has a `setFormClass()` method.

rudiedirkx:
And your tests :heart_eyes: could test for that:

```
some_container_injected_function(Kris\FormBuilder $forms1, MyFormBuilder $forms2) {
  $forms1 === $forms2 === app('kris-form-builder');
}
```

rudiedirkx (changes requested):
- Call the configs `form_helper_class`, `form_builder_class` (and `form_class`? if you want)
    - Use `form_class` to call form builder's `setFormClass()` from the SP. Or this could be another PR.
- Add **both** aliases if the class was changed to something else, not just the new alias

rudiedirkx (changes requested):
Test `it_register()` doesn't test anything. All the tested binds are already bound in `$this->app`, even without the `$provider` bit, so you could just remove that and it'd still pass.

rudiedirkx:
This `FormBuilderServiceProviderTest` is more useful. It overrides the `form_builder_class` config and tests if all the dependency injection works, with standard and custom abstracts.

ukeloop:
Thanks for fixing my codes.

rudiedirkx:
This code will have your name attached to it too ;) Are you okay with the current branch?

ukeloop:
Of course. Thanks for reviewing and fixing my PR.

PR #664, opened 2022-01-18
Refactored the `FormHelper` Class

ukeloop:
Refactored the `FormHelper` class. I didn't add type hints to avoid losing compatibility.


rudiedirkx:
Most of this doesn't do anything. What does `join` vs `implode` do? Are you just copying your IDE's improvements?

ukeloop:
`join` is an alias of `implode`. However, aliases are not a recommended function because they are for backward compatibility and may be obsolete.

This PR is for refactoring codes. it include no changed functions.

PR #665, opened 2022-01-18
Refactored the `Rules` Class

ukeloop:
Refactored the `Rules` class. I didn't add type hints to avoid losing compatibility.

### DIFF
--- a/src/Kris/LaravelFormBuilder/Rules.php
+++ b/src/Kris/LaravelFormBuilder/Rules.php
@@ -4,7 +4,6 @@ namespace Kris\LaravelFormBuilder;
 
 class Rules
 {
-
     /**
      * @var string|null
      */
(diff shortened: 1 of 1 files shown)


rudiedirkx:
I like it, BUT:

- why am I not seeing any Github tests?
- what do you mean type hints vs compatibility? PHP return types? ([We don't call them _hints_ anymore.](https://www.php.net/manual/en/language.types.declarations.php))
- most of the diff is whitespace..? Except for the `InvalidArgumentException`, that's a genuine bugfix. Is that the reason for this PR?

ukeloop:
> why am I not seeing any Github tests?

I'm sorry. I don't know that. Maybe it need authentications?

> what do you mean type hints vs compatibility? PHP return types? (We don't call them hints anymore.)

Yes. It's type declarations of return value and args. But it's a disruptive change. Compatibility will be lost if you are using custom fields.

I want type declarations. But I think it need interfaces for this project.

> most of the diff is whitespace..?

I think it's important to have a consistent code style within a project. I think this is even more necessary for custom fields, since you need to read the project code in order to create them.

> Except for the InvalidArgumentException, that's a genuine bugfix. Is that the reason for this PR?

This PR is for refactoring codes, not for bugfix.

ukeloop:
I pushed codes to trigger CI.

PR #670, opened 2022-02-26
collection input using empty_model

rudiedirkx:
If there's no original data, only `empty_row` data should be used, so only 1 row, AND formatted into models if `empty_model` is set.

### DIFF
--- a/src/Kris/LaravelFormBuilder/Fields/CollectionType.php
+++ b/src/Kris/LaravelFormBuilder/Fields/CollectionType.php
@@ -97,11 +97,19 @@ class CollectionType extends ParentType
 
         // If no value is provided, get values from current request.
         if (!is_null($data) && count($data) === 0) {
-            $data = $currentInput;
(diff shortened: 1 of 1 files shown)


rudiedirkx:
If the collection has 0 rows, AND 'empty_row` is true:

- only use the 1st input row, not every input row
- use `formatInputIntoModels` to format input into wanted model

rudiedirkx:
Crap. Maybe that's wrong logic. What if `prefer_input` is true, AND `empty_row` is true, and there's no input/database data? Then only `empty_row` is used, and input is cut off at index 1, so only 1 row, even if `prefer_input` is true to allow all rows.

rudiedirkx:
This is better. Accepted data now depends on `empty_model` and `prefer_input`. @kristijanhusak Objections?

kristijanhusak:
Looks ok to me.

kristijanhusak (approved):

PR #671, opened 2022-02-26
Drop Laravel less than 5.4

ukeloop:
This project doesn't support laravelcollective/html less than 5.4. 

https://github.com/kristijanhusak/laravel-form-builder/blob/ad01f1cf8eb2dd953fc4d4c9d3fa51d703d61723/composer.json#L14

### DIFF
--- a/src/Kris/LaravelFormBuilder/FormBuilderServiceProvider.php
+++ b/src/Kris/LaravelFormBuilder/FormBuilderServiceProvider.php
@@ -6,8 +6,6 @@ use Illuminate\Foundation\AliasLoader;
 use Collective\Html\FormBuilder as LaravelForm;
 use Collective\Html\HtmlBuilder;
 use Illuminate\Support\ServiceProvider;
-use Illuminate\Foundation\Application;
-use Illuminate\Support\Str;
 use Kris\LaravelFormBuilder\Traits\ValidatesWhenResolved;
 use Kris\LaravelFormBuilder\Form;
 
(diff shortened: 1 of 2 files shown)


rudiedirkx:
I don't even know why this package registers `'form'`. `laravelcollective/html` does that, no? The order doesn't matter, because it's `register()`. Why not completely remove `registerFormIfHeeded()` and `registerHtmlIfHeeded()`?

ukeloop:
`laravelcollective/html` is used in views.

rudiedirkx:
> `laravelcollective/html` is used in views.

And it's a dependency, so why register it twice? `laravelcollective/html` already register it, right?

rudiedirkx:
Tests fail without it, but that's not a reason to override a dependency's registration, AND it's easily fixable by extending `FormBuilderTestCase` to include a bit of `laravelcollective/html`:

```php
    protected function getPackageProviders($app)
    {
        return [
            'Collective\Html\HtmlServiceProvider',
            'Kris\LaravelFormBuilder\FormBuilderServiceProvider',
        ];
    }

    protected function getPackageAliases($app)
    {
        return [
            'Acme' => 'Kris\LaravelFormBuilder\Facades\FormBuilder',
            'Form' => 'Collective\Html\FormFacade',
            'Html' => 'Collective\Html\HtmlFacade',
        ];
    }
```

And then we can completely remove `registerHtmlIfNeeded()` and `registerFormIfHeeded()`.

rudiedirkx (changes requested):
Add `laravelcollective/html`'s service provider and facade to the test suite, and completely remove `registerFormIfHeeded()` and `registerHtmlIfHeeded()`.

ukeloop:
Thanks for fixing my codes. I must have missed the settings for package auto discovery.

Maybe we don't need this setup?

https://github.com/kristijanhusak/laravel-form-builder/blob/a7a93e6d78a2a9f0634b9bafb10313ec0310ce68/README.md?plain=1#L51-L68

rudiedirkx:
My 'fix' was only for testing. You're also correct in your quote, but that's yet another issue IMO =)

If you're okay with the current code, I'm gonna merge it, because I like it.

ukeloop:
Sure, please.

PR #674, opened 2022-03-05
Added `isPlain()` method for modified plain form

ukeloop:
Added `isPlain()` method for modified plain form.

```php

class ModifiedPlainForm extends Form
{
    protected function isPlain()
    {
        return static::class === self::class;
    }
}
```

```php
$formBuilder->setFormClass(ModifiedPlainForm::class);
```

### DIFF
--- a/src/Kris/LaravelFormBuilder/Form.php
+++ b/src/Kris/LaravelFormBuilder/Form.php
@@ -187,7 +187,11 @@ class Form
      */
     protected function isPlain()
     {
-        return static::class === self::class;
+        if($this->formBuilder === null) {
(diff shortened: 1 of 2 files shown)


rudiedirkx:
You shouldn't have to override `isPlain` if you want a custom plain form. `get_class($this) === 'Kris\LaravelFormBuilder\Form'` IS broken though, I agree. It should check against `$formBuilder->plainFormClass`. I think that's enough fix.

ukeloop:
> You shouldn't have to override `isPlain` if you want a custom plain form.

I fixed codes. Thx.

rudiedirkx:
Exellent. Great work this week @ukeloop !

PR #689, opened 2022-05-05
Fix deprecation message if str_replace receive a null value in FormHelper

ajira86:
> str_replace(): Passing null to parameter #3 ($subject) of type array|string is deprecated

Today str_replace is returning a blank string if a null value is specified as subject

### DIFF
--- a/src/Kris/LaravelFormBuilder/FormHelper.php
+++ b/src/Kris/LaravelFormBuilder/FormHelper.php
@@ -456,7 +456,7 @@ class FormHelper
      */
     public function transformToDotSyntax($string)
     {
-        if (!$string) {
+        if ($string === null) {
             return '';
         }
 
--- a/src/Kris/LaravelFormBuilder/FormHelper.php
+++ b/src/Kris/LaravelFormBuilder/FormHelper.php
@@ -456,6 +456,10 @@ class FormHelper
      */
     public function transformToDotSyntax($string)
     {
+        if (!$string) {
+            return '';
+        }
+
         return str_replace(['.', '[]', '[', ']'], ['_', '', '.', ''], $string);
     }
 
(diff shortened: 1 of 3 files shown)


src/Kris/LaravelFormBuilder/FormHelper.php
```diff
@@ -456,6 +456,10 @@ public function appendMessagesWithPrefix(MessageBag $messageBag, $prefix, array
      */
     public function transformToDotSyntax($string)
     {
+        if (!$string) {
```
rudiedirkx, line 459: That also triggers for `0` and `'0'`. Probably never happens, but still.

rudiedirkx (changes requested):

src/Kris/LaravelFormBuilder/FormHelper.php
ajira86: Oh, I followed the code style from line 277 but yeah better to check with is_null method.

ajira86:
> But how/why/where is `transformToDotSyntax()` ever called with `null`?

I'll check what happens, I'm curious too ^^

ajira86:
So, it happens because `getValidationRules` is calling `getNameKey` from the `Form`.

In my case, my forms are not named, this is why a null value is passed to `transformToDotSyntax`

rudiedirkx:
In that case `Form::getNameKey()` should probably check for `name` nullness. IMO `transformToDotSyntax` shouldn't accept `null`, but we can't change that now, so adding a null-check is still smart.

But why doesn't any test break on this? We have PHP 8.1 tests... With nameless forms. Maybe deprecateds don't fail tests, hmm...

rudiedirkx:
I've fixed the test suite. Deprecations are now errors. And I left another one for you ;)

(I merged that into your branch, so you'll have to pull your refactor branch.)

rudiedirkx:
Thank you for the bugfind, and the realization that deprecations were ignored in tests. You're testing the form builder well in PHP 8.1? I'm not at 8.1 yet.

ajira86:
Is it possible to publish a version with this fix ?

rudiedirkx:
Done. You should probably just use `dev-master`, because we're bad at releases, and you have to keep an eye on changes anyway.

PR #693, opened 2022-05-15
Unify return values

ukeloop:


rudiedirkx:
What's unifying about this? Why does it help? I don't get it.

rudiedirkx:
I was paying poor attention. Apologies. The return values make complete sense now that I'm actually reading :blush: Get in!

PR #698, opened 2022-10-15
accept closure attributes (eg wrapper.class)

rudiedirkx:
This change accepts attributes to be closures (resolved by Laravel's `value()`). That's not useful for most fields, because all context is present where the field is defined. But it is useful for `collection` fields, because you define the same attributes for all the items/child forms. This way you can dynamically add different attributes for different items:

```

class ParentForm {
	function buildForm() {
		$this->add('items', 'collection', [
			'type' => 'form',
			'prototype' => false,
			'empty_model' => fn() => new ChildModel,
			'options' => [
				'class' => ChildForm::class,
				'wrapper' => [
					// current
					'class' => 'non-dynamic class',
					// new, optional
					'class' => function(\Kris\LaravelFormBuilder\Fields\FormField $field) {
						return 'use $field to add dynamic class';
					},
				],
			],
		]);
	}
}
```

I use this to highlight some collection items if they contain problems.

`$field` is the collection item field. In this case a `ChildFormType` because `type = form`. It could be a `InputType`, which doesn't have a `getForm(I`

### DIFF
--- a/src/Kris/LaravelFormBuilder/Fields/FormField.php
+++ b/src/Kris/LaravelFormBuilder/Fields/FormField.php
@@ -323,13 +323,13 @@ abstract class FormField
             }
         }
 
-        $this->setOption('wrapperAttrs', $helper->prepareAttributes($this->getOption('wrapper')));
-        $this->setOption('errorAttrs', $helper->prepareAttributes($this->getOption('errors')));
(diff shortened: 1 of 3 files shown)


Eloar:
Why it is not yet merged? It seems to be neat one to be introduced

rudiedirkx:
Because I didn't need it after all, and change is scary.

PR #700, opened 2022-11-30
label_template option

Eloar:
new option for label template and moved rendering label to `label.php` template file for all templates with label. It simplifies extension of way label is rendered and allows to add custom html to labels. Works for every field (by extending `label.php` file) or for individual field (by providing `label_template` value pointing proper template file).

### DIFF
--- a/tests/FormBuilderTestCase.php
+++ b/tests/FormBuilderTestCase.php
@@ -185,6 +185,8 @@ abstract class FormBuilderTestCase extends TestCase {
         parent::setUp();
 
         $this->withoutDeprecationHandling();
+        // add views for testing
+        $this->app['view']->addNamespace('laravel-form-builder-test', __DIR__ . '/resources/views');
 
         $this->view = $this->app['view'];
         $this->translator = $this->app['translator'];
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -14,6 +14,9 @@
     <testsuites>
         <testsuite name="Package Test Suite">
             <directory suffix=".php">./tests/</directory>
+            <exclude>
+                <directory>./tests/resources/views</directory>
+            </exclude>
         </testsuite>
     </testsuites>
     <coverage>
--- a/src/helpers.php
+++ b/src/helpers.php
@@ -37,6 +37,14 @@ if(!function_exists('helpBlockPath'))
 
 }
 
+if (!function_exists('labelBlockPath'))
+{
+    function labelBlockPath()
+    {
+        return getFormBuilderViewPath('label.php');
+    }
+}
+
 if (!function_exists('form')) {
 
     function form(Form $form, array $options = [])
--- a/src/views/checkbox.php
+++ b/src/views/checkbox.php
@@ -7,8 +7,13 @@
 <?php if ($showField): ?>
     <?= Form::checkbox($name, $options['value'], $options['checked'], $options['attr']) ?>
 
+    <?php /** label rendering section */ ?>
     <?php if ($showLabel && $options['label'] !== false && $options['label_show']): ?>
-        <?= Form::customLabel($name, $options['label'], $options['label_attr']) ?>
+        <?php if(array_key_exists('label_template', $options) && $options['label_template']): ?>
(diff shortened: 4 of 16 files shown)


rudiedirkx:
I like it, BUT:

- You've committed some merge conflicts.
- And white-space is messy. Use indents like you do in `checkbox.php`.
- But more importantly, the fallback should be done like the help & error block:

```
<?php include helpBlockPath(); ?>
<?php include errorBlockPath(); ?>
```

instead of

```
<?php include 'label.php' ?>
```

so the dev can override the label in `views/vendor/laravel-form-builder/` for all labels always.

- And I feel like the 'if should show label' should be outside the label view:

```
<?php if ($showLabel && $options['label'] !== false && $options['label_show']): ?>
```

not inside it, but maybe I'm misunderstanding your goals. Can you give an example of a label override view? What's in there?

rudiedirkx (changes requested):
That's getting good, buuutt now we're repeating this exact logic A LOT:

```php
<?php /** label rendering section */ ?>
<?php if ($showLabel && $options['label'] !== false && $options['label_show']): ?>
    <?php if(array_key_exists('label_template', $options) && $options['label_template']): ?>
        <?= view($options['label_template'], get_defined_vars())->render(); ?>
    <?php else: ?>
        <?php include labelBlockPath(); ?>
    <?php endif; ?>
<?php endif; ?>
```

Maybe we can create a form builder helper for that? Something that takes `get_defined_vars()` and does the same? A static function in `FormHelper`?, or is that evil?

Eloar:
Thought about some solution to reduce code repetition but there is plenty of it in form elements templates already. Like this exact code is repeated 12 times:

```php
<?php if ($options['wrapper'] !== false): ?>
<div <?= $options['wrapperAttrs'] ?> >
<?php endif; ?>
```

Not sure if new repetition becomes a problem as new helper methods or functions would be repeated just as much with this logic hidden inside. Although changing label visibility condition would need change in one file not 9. On the other hand changing that repeated code just for one form element type would be easier for package user as this logic would not be hidden in helper. I am bit conflicted regardless of that idea. 

Echo short tags (`<?= ?>`) are available even when `short_open_tag` setting is turned off on PHP interpreter. I believe it is so since PHP 5.4.  It is described on [echo page in docs](https://www.php.net/manual/en/function.echo.php).

tests/FormBuilderTestCase.php
```diff
@@ -184,7 +184,9 @@ public function setUp(): void
     {
         parent::setUp();
 
-        $this->withoutDeprecationHandling();
+        // $this->withoutDeprecationHandling();
```
rudiedirkx, line 187: Why? Deprecations are errors we should find, not silently ignore.

rudiedirkx:
Well damn Github. I wrote a reply and Github threw it away.

----

short open tags: you're right, and we already use them, so ignore me.

repetition: I don't really mind either. Let's not add more helper functions.

rudiedirkx (changes requested):
Aside from the `withoutDeprecationHandling()` I'm happy.

rudiedirkx:
You keep changing tests. Why? They work for me and for Github Actions. Is your test setup weird, or mine? Apparently they don't play nice together, because your exception change doesn't work.

rudiedirkx:
(You don't have to squash and rebase on your end. Github will do that when merging. That's easier to track PR progress.)

Eloar:
What makes my setup weird is I started from 1.22 version as it is last working with Laravel 5.5 (at least as I confirmed). I have legacy project in 5.5 that I want to use that feature in and it means I'm switching to my fork and using 1.23 tag. I started from 1.22 and fitted it to master which introduced trouble for pull request. Should have started other way around.

Fixed what I messed up in test so it should work now.

rudiedirkx:
You've committed a few cache files that should be in your `.gitignore`. We'll get there eventually =)

Eloar:
You are absolutely right, that should stay out of repo. Modified and pushed. Damn I'm clumsy with this one.

rudiedirkx:
I think we're done :+1: And if not, bring the reverts!

PR #705, opened 2023-02-13
ParentType  disable()/enable() fix

Eloar:
Fields extending `ParentType` don't act properly when disabled. For example `ChoiceType` with `required` rule is not disabled even if asked nicely with `disable()` method. This is due to fact `attr` option gets copied to children from parent while rendering but it doesn't contain `disabled` attribute after `disable()` method was called.

Fixed `disable()` and `enable()` method to act upon `ParentType` field itself before populating information to children. Added functionality tests to all derivatives (as ParentType is abstract and has no tests written so far).

Don't know why it ads some code from previous PR that was already merged. Those changes were made during Pull Request review.

### DIFF
--- a/src/Kris/LaravelFormBuilder/Fields/ChoiceType.php
+++ b/src/Kris/LaravelFormBuilder/Fields/ChoiceType.php
@@ -2,6 +2,8 @@
 
 namespace  Kris\LaravelFormBuilder\Fields;
 
+use Illuminate\Support\Arr;
+
 class ChoiceType extends ParentType
 {
     /**
@@ -32,7 +34,7 @@ class ChoiceType extends ParentType
         $expanded = $this->options['expanded'];
         $multiple = $this->options['multiple'];
 
-        if ($multiple) {
+        if (!$expanded && $multiple) {
             $this->options['attr']['multiple'] = true;
         }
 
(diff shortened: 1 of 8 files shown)


rudiedirkx:
I dislike `ParentType`. I don't use `ChoiceType`, but [rdx/laravel-form-builder-extras](https://github.com/rudiedirkx/laravel-form-builder-extras), because it's so broken in many ways. I'll check out the PR soon, but 23 files changed.... not a good sign.

Eloar:
I don't understand why my forks master contains 2 commits from label_template feature that were not merged yet pull request was. I use ChoiceType which means dealing with ParentType. I know it has some problems and I finally found a time to fix one that bothered me from time to time.

rudiedirkx:
Can you remake the diff/PR?, because there's a lot of different stuff in here. That will be confusing in the future.

And can you explain the problem this fixes? Does that problem only happen if you use

Eloar:
I rebased branch to keep just fix commit here and those left commits for `label_template` option that should be merged with #700 on Dec 17 2022 I will add as separete PR.

rudiedirkx:
And can you explain the problem this fixes? Does that problem only happen if you use `$field->enable/disable()`? (I never do, so I've never seen it.)

Eloar:
I experience it using ChoiceType field with `required` rule set on it and form disabled by calling `disable()` method on it. Some debugging let me found out that not ChoiceType but ParrentType is a problem and certine options are not properly populated from parent to children. This way when whole form is set as disabled ParentType derivative fields stay enabled.

rudiedirkx:
Oh man I really don't like (or understand anymore) `ParentType`. I'm sure this PR fixes your problems, but what about other effects? I only use `CollectionType`, not `ChoiceType`, so I can't test this in real code. How confident are your this won't break anything existing? I only see _more_ attribute/option copying to children, not less, and I thought all the copying was the problem. I understand copying `disable` and `enable`, but _all_ attributes/options??

Eloar:
All tests checks out so I would assume it works without breaking those types beyond current state. I use forked version with this fix for some time and have no complaints for neither `Choice` nor `Collection` type fields on my forms.

This was done long enough so I'm not sure but it seems some other changes made on already built form are ignored for parent based fields without it. This was a reason why more then 'disabled' attribute is copied.

rudiedirkx:
Alright, let's do it. I can always blame you when someone complains ;)

PR #706, opened 2023-02-15
Support Laravel 10

vigstudio:
Support Laravel 10

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,9 +11,9 @@
     ],
     "require": {
         "php": ">=7.1",
-        "laravelcollective/html": "^5.6|^6|^7|^8|^9",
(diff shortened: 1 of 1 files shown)


rudiedirkx:
Tests install laravel 8.83.27, not 10. Why? Other dependencies?

vigstudio:
> Tests install laravel 8.83.27, not 10. Why? Other dependencies?

image
because version 8 is latest and it supports laravel 10

rudiedirkx:
testbench 8.x breaks all kinds of stuff, so I've included 7.x. See master. Tests run with laravel 9 now. Still not great, but I don't feel like fixing unit tests that I know nothing about. PR for testbench 8 + laravel 10 + newest phpunit very welcome.

PR #709, opened 2023-04-12
Label template

Eloar:
Continuation to #700 that somehow got left after review. Don't know how it was not merged on Dec 17 2022 as those commits were made earlier. All those changes were made according to requests from #700 PR.


Eloar:
On one hand it seems like so, on the other github still reports changes here to master and it gets confusing.

rudiedirkx:
That's because it was squash merged I think. No need for this PR. Look at the code in `master`. It's all there.

Issue #447, opened 2018-08-03
Form is valid even if repeated fields are not the same


vince83110:
I use Laravel 5.6 and your package, but I can't show an error when repeated fields are not the same.

I have 2 repeated fields in my form (email and password), I put different passwords but 

`$form->isValid()`

returns true. Do I need to write my own validation ? 

My password form is : 

```
            ])->add('password', 'repeated', [
                'type' => 'password',
                'second_name' => 'password_confirmation',
                'first_options' => [
                    'template' => 'layout.forms.text_right',
                    'rules' => 'required',
                    'label' => __('customer.password'),
                ],
                'second_options' => [
                    'template' => 'layout.forms.text_right',
                    'rules' => 'required',
                    'label' => __('customer.password_confirmation'),
                ],
            ])
```

Thanks

kristijanhusak:
Yeah, you need to write your own validation, there's no default support for that.
This field type was added before the validation feature, so we should maybe update it to support that.

PR #710, opened 2023-04-13
refs #447: adding `same:` rule automatically to repeated field type

Eloar:
- adding `same:` rule to first field of repeated type
- proper handling of rules on repeated field itself and first subfield

resolves issue #447


rudiedirkx:
Great fix! Another `ParentType` that I never use, but at least I understand this one =) Thanks for the very overdue fix.

Issue #677, opened 2022-03-14
Bugs when modding multiple times with `multiple`


ukeloop:
Bugs when modding multiple times with `multiple`.

https://github.com/kristijanhusak/laravel-form-builder/issues/305 related bugs.

```php
$options = [
    'choices' => ['en' => 'English', 'fr' => 'French'],
    'choice_options' => [
        'wrapper' => ['class' => 'choice-wrapper'],
        'label_attr' => ['class' => 'label-class'],
    ],
    'expanded' => true,
    'multiple' => true
];

$this->plainForm->add('test', 'choice', $options);

echo $this->plainForm->test->getName(); // test
var_dump($this->plainForm->test->getOption('tmp.multipleBracesSet')); // Null

$this->plainForm->modify('test', 'choice', [
    'rules' => 'required',
]);

echo $this->plainForm->test->getName(); // test[]
var_dump($this->plainForm->test->getOption('tmp.multipleBracesSet')); // true

$this->plainForm->modify('test', 'choice', [
    'rules' => 'required',
]);

echo $this->plainForm->test->getName(); // test
var_dump($this->plainForm->test->getOption('tmp.multipleBracesSet')); // true
```

I think codes of modding name should move to `render` method and remove side effects.

https://github.com/kristijanhusak/laravel-form-builder/blob/735c258e3ac3e524226eee5d898aec07e4812594/src/Kris/LaravelFormBuilder/Fields/FormField.php#L293-L296

rudiedirkx:
The `choice` type is broken is several ways. All `ParentType` fields are somewhat broken. I'm starting v2 without it.

Issue #324, opened 2017-02-14
Multiple checkboxes in plain subform get double `[][]`


rudiedirkx:
My form simplified:

````
// Prepare subform
$taco_test = $this->formBuilder->plain();
$taco_test->add('professions', 'choice', [
  'label' => trans('report.taco_test.professions'),
  'choices' => $favorites_options,
  'multiple' => TRUE,
  'expanded' => TRUE,
]);

// Add subform to parent
$this->add('taco_test', 'form', [
  'label' => trans('report.taco_test.taco_test'),
  'class' => $taco_test, // the plain form object
]);
````

So the checkboxes live at `form.taco_test.professions`, so a single checkbox should be at eg. `form.taco_test.professions.66`.

The resulting checkboxes have `name="taco_test[professions][][]"`, with a double `[][]`. Their values are correct.

I know the last `[]` are added in `ChoiceType::buildCheckableChildren()`, but maybe it's the first `[]` that's actually wrong. I've no idea where that comes from.

I'm using 1.6.50 and I'm having some dependency problems, so I can't update in this context to try newer version, but the `ChoiceType` code seems to be the same in HEAD.

**Comment by rudiedirkx** [comment]
If I `dd($taco_test);` before adding it to the parent form, I can see all checkboxes' names:

    #name: "professions[]"

That `[]` comes from `buildCheckableChildren()` and I think it's correct, so it seems another is added after it...

**Comment by rudiedirkx** [comment]
Temporary local 'fix':

```
  public function buildForm() {
    // ...
    $this->fixChildFormCheckboxes($this->taco_test->professions);
  }

  protected function fixChildFormCheckboxes(FormField $field) {
    foreach ($field->getChildren() as $child) {
      $child->setName($field->getName());
    }
  }
```

because `form.taco_test.professions`'s name is correct: `taco_test[professions][]`. The children shouldn't all get an extra `[]`.

rudiedirkx:
Still broken, but too complex for me. Work-around: explicit `setName()` after every plain form creation:

    $career_choice_test = $this->formBuilder->plain()->setName('career_choice_test');

and in a subform:

    'class' => $this->formBuilder->plain()->setName("career_choice_test[$name]")

Issue #305, opened 2017-01-18
Checkbox get two "[][]" if using modify


SebTM:
Hello,

if I use modify on field-type "choice" I get html like "fieldName[][]" as name attribute. Here is my controller-code to reproduce:

```
public function testModify()
    {
        $form = FormBuilder::plain()
            ->add('name', 'text')
            ->add('test', 'choice', [
                'choices' => ['en' => 'English', 'fr' => 'French'],
                'choice_options' => [
                    'wrapper' => ['class' => 'choice-wrapper'],
                    'label_attr' => ['class' => 'label-class'],
                ],
                'expanded' => true,
                'multiple' => true
            ]);

        $form->modify('test', 'choice', [
            'rules' => 'required',
        ]);

        $form->add('submit', 'submit');

        return View::make('cdou.form', ['form' => $form]);
    }
```

Best Regards!

kristijanhusak:
@SebTM please avoid `modify` method, since there are some issues with it https://github.com/kristijanhusak/laravel-form-builder/issues/237

SebTM:
Hello,

I understand that there are currently issues with "modify". I really need to use the method as this example is only to reproduce the issue. I have a complex crud-system which uses form-classes to provide the basic-functionallity but the validation-configuration comes from another configuration.

Is there any possibility to archiv/fix this?

yemenifree:
@SebTM I have the same problem, as quick fix pass "tmp" to options field to avoid add [] multi times

` 
'tmp' => ['multipleBracesSet' => true],
`

PR #715, opened 2023-09-13
Fix bugs in ChoiceType's name

ukeloop:
Fix bugs in `ChoiceType`'s name.

## Related issues

- https://github.com/kristijanhusak/laravel-form-builder/issues/677
- https://github.com/kristijanhusak/laravel-form-builder/issues/324
- https://github.com/kristijanhusak/laravel-form-builder/issues/305

### DIFF
--- a/src/Kris/LaravelFormBuilder/Fields/ChoiceType.php
+++ b/src/Kris/LaravelFormBuilder/Fields/ChoiceType.php
@@ -1,6 +1,6 @@
 <?php
 
-namespace  Kris\LaravelFormBuilder\Fields;
+namespace Kris\LaravelFormBuilder\Fields;
 
 use Illuminate\Support\Arr;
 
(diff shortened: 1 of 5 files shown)


rudiedirkx:
Yeeaah, but `tmp` is so ugly... =( I rather just get rid of `ChoiceType` and `ParentType` and create a breaking 2.x with a very simple upgrade path. It's **always** problems with these fields 😠 

I'm okay with this PR, because I don't use `ChoiceType`, so I know it definitely won't break my code, but I'm not sure enough about ChoiceType ever.

rudiedirkx:
LOL what a complainer #324 :D

ukeloop:
@Eloar 

See that:

https://github.com/kristijanhusak/laravel-form-builder/blob/d8c2ba682bbfc965ee10da2eeec54627a22c2a8a/src/Kris/LaravelFormBuilder/Fields/FormField.php#L304-L307

Issue #719, opened 2024-02-26
Invalid pattern for input


Eloar:
Noticed multiple errors in a browser console for my form. Apparently pattern field on input is invalid. It contains pattern `[-+]?[0-9]*[.,]?[0-9]+`

PR #720, opened 2024-02-26
issue 719: fixed numeric pattern

Eloar:
fix for #719, simple escape for invalid character.


Eloar:
`\` escaped. I don't necessarily see it as an improvement in readability, but won't argue against that change.

rudiedirkx:
In json it's a mandatory escape, in regex too (if you want to match a literal `\`), only in PHP it's not. I like to keep them similar. Anyway, good catch, thanks 👍

PR #721, opened 2024-02-27
Laravel 11.x Compatibility

laravel-shift:
This is an automated pull request from [Shift](https://laravelshift.com) to update your package code and dependencies to be compatible with Laravel 11.x.

**Before merging**, you need to:

- Checkout the `l11-compatibility` branch
- Review **all** comments for additional changes
- Thoroughly test your package

If you do find an issue, please report it by commenting on this PR to help improve future automation.

PR #724, opened 2024-03-26
Update composer.json to support Laravel 11

thanosalexandris:


rudiedirkx:
#721 Are you sure it works well in L11? Discussion there.

PR #725, opened 2024-04-03
Registering customLabel macro once Form resolved

sangnguyenplus:


rudiedirkx:
Why? Service provider order? That has never been a problem for me. Are your SP ordered differently/manually?

sangnguyenplus:
@rudiedirkx Sorry, I forgot to explain it.

I have changed to use `afterResolving(static::FORM_ABSTRACT, (the form service)`

Actually, you still can use it, and it won't cause any bugs, I have used your package many years and it is still working fine.

But today, I checked this code when trying to replace `laravelcollective/html` package (it's deprecated already).

**laravelcollective/html/src/HtmlServiceProvider.php:59**

image

When using `form` service, it will initialize session, so if you put that code directly in boot() method without wrapping it in afterResolving(static::FORM_ABSTRACT), it will initialize session in boot method, it's not correct app lifecycle. Many HTTP requests don't need session, so `session` shouldn't be resolved in boot (the same with artisan commands).

With this change, `session` won't be started in boot() method of service provider.

rudiedirkx:
I require [`rdx/laravelcollective-html`](https://packagist.org/packages/rdx/laravelcollective-html) in my projects, which composer-replaces `laravelcollective/html` for compatibility. This package still requires `laravelcollective/html` to not force a replacement. Such a shame that they quit like this =( What a mess it is now.

How are you replacing `laravelcollective/html`?

rudiedirkx:
If you're feeling extra generous, you could look at the tests for Laravel 11 support #721 😄 I don't understand phpunit...

sangnguyenplus:
@rudiedirkx It just has a few classes https://github.com/LaravelCollective/html/tree/6.x/src. I have copied it into our core project and maintain it.

PR #726, opened 2024-04-05
Revert #725 - Registering customLabel macro once Form resolved

sangnguyenplus:
#725

@rudiedirkx It is working fine in our latest code with Laravel 10, but I got some reports that it doesn't work for old versions (which using Laravel 9).

image

I think this is a breaking change, we should not change this at this time.

Please revert it.

### DIFF
--- a/src/Kris/LaravelFormBuilder/FormBuilderServiceProvider.php
+++ b/src/Kris/LaravelFormBuilder/FormBuilderServiceProvider.php
@@ -7,8 +7,8 @@ use Collective\Html\FormBuilder as LaravelForm;
 use Collective\Html\HtmlBuilder;
 use Illuminate\Support\ServiceProvider;
 use InvalidArgumentException;
-use Kris\LaravelFormBuilder\Form;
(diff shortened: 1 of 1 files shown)


rudiedirkx:
Works for me with Laravel 9 on PHP 8.2. Original PR tested Laravel 9 on PHP 8.0, that worked too: https://github.com/kristijanhusak/laravel-form-builder/actions/runs/8547107396/job/23418677886?pr=725

What seems to be going wrong? Any idea? I liked the smarted boot, I don't want to revert it without a good specific known reason.

sangnguyenplus:
@rudiedirkx I have no idea, if it is fine for you then we may keep it. Maybe it has something wrong with their code, not an issue of this package.

rudiedirkx:
If you get more details or reports, add them here. Maybe our tests just aren't good enough, but the `afterResolving` change seems good to me.

rudiedirkx:
@sangnguyenplus 1.53.1 probably fixes this.

